Issue #58, opened 2026-07-06
Configurable resize step (global `resize.step` setting)


hajiboy95:
**Spun off from #23 (Wave 6).** The keybinding catalog now ships
**Grow / Shrink** presets that emit `KiwiDesk.resize("x", ±50)` at a
**fixed 50 pt step**. Make that step configurable.

**Why a global setting, not a per-row slider.** The Grow/Shrink
rows are presets keyed by their *exact* Lua, and the #4 import
classifier matches on that exact string. A per-row slider that
rewrites the literal (`resize("x", 30)`) would change the row's
identity key *and* break exact-match import. The clean design is a
single global step:

- New tiling setting `resize.step` (per the one-vocabulary
  guardrail — JSON key derived from the Lua name via `CodingKeys`).
- New Lua command `KiwiDesk.set_resize_step(50)`.
- Grow/Shrink reference the step; a GUI slider/stepper (e.g.
  10–200 pt) edits it.
- Import classification then matches on resize *shape*
  (`resize("x", ±N)`) and reads `N` back into the step, instead of
  the current exact-Lua match.

**Scope:** Core setting + command, config writer, `SettingsCodingTests`
(pins the JSON shape), docs. Small but real — its own PR.

**Related:** #56 (true 2-axis resize would extend this to per-axis
steps).

hajiboy95:
Flagged in the Wave 8 redesign capstone (#68, "Must keep in mind"): the new Settings IA should reserve room for the `resize.step` control next to Grow/Shrink (and for per-axis steps once #56 lands). The setting/command/import work here stays its own PR — the redesign only re-presents it.

Issue #56, opened 2026-07-06
resize: true 2-axis (independent H/V) resizing


hajiboy95:
## Problem

`KiwiDesk.resize(axis, delta)` accepts `"x"`/`"y"`, which *looks*
like independent 2-axis resizing — but the current layout model
can't deliver it:

- **BSP** is dwindle-style over a flat array with a **single
  `splitRatio` scalar** (`BspLayout.swift`): every split at every
  depth — horizontal *and* vertical — uses that one ratio.
- **Stack** likewise has a single `masterRatio`.
- In `resize` (`KiwiCore+Commands.swift:147-163`), **both `"x"` and
  `"y"` write that same scalar.** The axis only changes the step
  scaling (`delta ÷ screenWidth` vs `÷ screenHeight`).
- **Scrolling ignores the axis entirely** — it resizes the slot
  along its own bar axis.
- **monocle / grid / floating**: `resize` returns "not supported".

Net: `resize("y")` does *not* independently control vertical size;
it nudges the same knob as `resize("x")`. In BSP the dominant first
split is usually horizontal, so a "resize vertically" key visibly
changes column *widths* — the wrong direction.

## Docs currently oversell this

- `docs/configuration.md:531-535` shows a resize mode mapping
  `h/l → resize("x", ∓50)` and `j/k → resize("y", ±50)` as if the
  two axes were independent. Misleading given the single ratio.
- `docs/cli.md:109-110` is accurate but silent on the axis
  limitation and the unsupported modes.

## Goal

Make `resize` genuinely 2-dimensional so `"x"` and `"y"` are
distinct behaviors — e.g. give BSP/Stack independent horizontal
and vertical ratios (or per-node ratios), and have Scrolling honor
the requested axis where it makes sense. Then a "resize vertically"
key actually resizes vertically.

## Acceptance

- [ ] `resize("x", …)` and `resize("y", …)` produce independent
      horizontal vs vertical changes in BSP and Stack.
- [ ] Decide and document behavior for Scrolling / monocle / grid /
      floating (support or explicit no-op).
- [ ] Correct `docs/configuration.md` (the resize-mode example) and
      `docs/cli.md` to describe the real per-axis behavior and the
      unsupported layouts.
- [ ] Upgrade the keybindings catalog (#23) from the 3 rows
      (Grow / Shrink / Make floating) to **5**: Grow width, Shrink
      width, Grow height, Shrink height, Make floating.
[...]

Issue #51, opened 2026-07-06
feat(profiles): persist animation duration per-profile


hajiboy95:
> Spun off from the animation-vocabulary regrouping on
> `feat/animations-toggles` (the change that made
> `animations.set_duration(ms)` the primary command alongside the
> five `animations.*` toggles). Tracked so we don't forget it.

### Use case

The five animation toggles now persist **per-profile**, but the
animation **duration** does not. `animations.set_duration(ms)` (and
its deprecated alias `set_animation_duration`) stays **runtime-only**
— it can only be set in `init.lua`, so it applies globally and can't
differ between profiles the way the toggles now can. A user who wants
snappier animations in one profile and slower ones in another has no
way to express that.

Duration was deliberately left runtime-only in that PR because
`scroll.set_speed` writes the **same** `durationMS` knob, and neither
has ever been persisted. Persisting one without reconciling the other
would produce a stale/conflicting value on profile load. This issue
tracks doing it properly.

### Proposed solution

Persist duration per-profile, deriving the JSON key from the Lua name
per the repo's one-vocabulary rule (`animations.set_duration` →
`animations.duration`). The real work is the reconciliation, not the
storage:

- Decide how a persisted `animations.duration` reconciles with
  `scroll.set_speed` (which mutates the same `durationMS`) — single
  source of truth, precedence, or unify the two knobs.
- Add sync-on-profile-apply so switching profiles updates the live
  `durationMS` (and stays consistent with whatever `scroll.set_speed`
  did at runtime).
- Round-trip tests (save → load → apply) pinning the reconciled value,
  alongside the existing `SettingsCodingTests` shape check.

### Alternatives considered

Keep duration runtime-only (current behavior) — set once in
`init.lua`, shared with `scroll.set_speed`. Simple and consistent, but
no per-profile durations.

### Extra context

Relates to the animation work in #11. The command surface and docs
already state plainly that duration is runtime-only and shared with
`scroll.set_speed`; this issue is the follow-up to make it
per-profile if we decide it's worth the reconciliation.

### Personal Update

Actually the `scroll.set_speed` should be under animations.scroll_speed. it's specific forthe scrolling layout. therefore it will also need it's own GUI Element for setting this one.

hajiboy95:
[...]

PR #88, opened 2026-07-08
feat(gui): Settings redesign — sidebar IA, quick menu, EditTarget (#68)

### LINKED ISSUE #33: Multiple keybinding recorders can be active at once
### Problem

In the Shortcuts tab you can put **multiple keybinding recorders
into "Press keys…" at the same time** — clicking Record on a second
row doesn't cancel the one already listening (e.g. Focus Down and
Focus Up both waiting for input at once). Only one recorder should
capture input at a time.

### Suspected cause

`KeyRecorderField`
(`Sources/KiwiDesk/Settings/Tabs/KeyRecorderField.swift`) keeps
`recording` as per-field local `@State` and installs its own
`NSEvent.addLocalMonitorForEvents` on `start()`; nothing coordinates
a single active recorder across rows. A shared "currently-recording"
owner (e.g. a token on `SettingsModel`) would let starting one
recorder stop the others.

### Done when

At most one recorder is in "Press keys…" at any moment; starting a
new recording stops any other and tears down its event monitor.

### LINKED ISSUE #34: Prevent duplicate shortcuts within a mode (inline feedback)
### Use case

Two actions in the **same mode** can currently be bound to the same
combo, but only one can ever fire — the other is silently shadowed.
Today a duplicate is accepted and only flagged after the fact (the
⚠️ mark + conflict banner from #10). It should be actively prevented
at input, with immediate feedback.

### Proposed solution

When a recorded combo is already bound to another action **in the
same mode**, reject the assignment (keep the field's previous value,
or leave it empty) instead of accepting a duplicate. Give immediate
feedback at the field: a brief red flash on the recorder plus an
inline message such as `Already assigned to "Focus Left"` next to or
below it.

### Alternatives considered

- **Allow-but-warn (today's behavior)** — too permissive; a
  silently-shadowed binding is a footgun.
- **Scope of hard-prevention** — only *intra-mode duplicates* are
  hard-blocked (genuinely invalid: two actions can't share one combo
  in a mode). **macOS-system conflicts stay warnings** (the existing
  banner), since deliberately overriding a system shortcut can be a
  legitimate choice.

### Extra context

Builds on the conflict detection from #10 — `KeybindingConflicts`
already produces "Already bound in this mode: <label>"; this promotes
that specific case from a passive warning to active prevention.
Relates to the single-active-recorder bug (#33) and the
keybinding-editor work (#23).

### Done when

Recording a combo already used in the same mode is rejected with
immediate inline feedback (red flash + "already assigned to
<action>"), and the combo is never silently duplicated; macOS
conflicts remain allowed-with-warning.

### LINKED ISSUE #35: Conflict banner doesn't update when a shortcut is deleted
### Problem

The in-app conflict banner (#10) refreshes/clears when you
**record** a shortcut, but not when you **delete or clear** one. If
you clear the combo that caused a conflict (the ✕ on the recorder,
or removing a custom row), the row's ⚠️ disappears but the banner
stays — it only catches up on the next record.

### Suspected cause

The set-or-clear evaluation (`noteRecordedCombo` → update
`keybindingWarning`) is wired into `KeyRecorderField.onRecord` only.
The clear/remove paths — `NavRow`'s clear, `CustomSection`'s clear
and its row `remove`, and `ApplicationsSection` — mutate the
combo/bindings without re-running the conflict check.

### Proposed fix

Run the same evaluation after **any** change to a binding's combo
(record OR clear/remove): recompute
`KeybindingConflicts.conflicts(config.modes)` and update
`keybindingWarning` (clear when none remain, refresh the list when
some do). Factor the trigger so add and remove share one path.

### Done when

Clearing or removing the shortcut that caused a conflict updates the
banner immediately — and clears it if that was the last conflict.

### LINKED ISSUE #64: refactor(settings): collapse the dashboard edit target into an explicit EditTarget enum
### Context

Spun off from the #18 review (edit stored profiles). The architect
review flagged this as the single highest-value structural follow-up.

`SettingsModel` now keys its edit mode off `editingProfile: String?`
(nil = live/active config; a name = a stored profile edited without
activating). Two independent `reload()` paths branch on it —
`reload()` (live) and `reloadEditingProfile(name)` — and each must
*remember* to reset every mode-dependent field. That is exactly how
the two round-1 review defects arose: the raw-Lua-editor leak
(`showLuaEditor` not cleared) and the stale `placementEditable` on
the live path. Both were patched in #18, but the shape still invites
the same class of bug for the next field added.

### Proposal

Replace the optional-String target with a total enum and a single
reset-everything reload:

```swift
enum EditTarget: Equatable {
    case live
    case storedProfile(String)
}
```

- One `func reload(for target: EditTarget)` (or a `didSet` on the
  target) that **always** assigns every mode-dependent field:
  `config`, `savedSidecar`, `forcedLuaEditor`, `showLuaEditor`,
  `placementEditable`. Exhaustive by construction — the two modes
  can no longer drift.
- Optionally fold the live sub-states (`activeProfile` /
  `activeStandard` / transient) into the same enum, or a sibling,
  so the banner's title/status derive from one value instead of
  three published fields (the architect also suggested collapsing
  `currentName` / `currentStandard` / `isDirty` in `ProfileManager`
  — evaluate together).
- Because the logic moves into the model, the mode transitions
  become unit-testable (today they are GUI-only and untested).

### Not in scope / related deferred cleanups

Two other opportunistic items noted in the #18 plan, independent of
this refactor:
- Unify the GUI→core seam: `SettingsModel.loadProfile` /
  `deleteProfile` / `makeDefault` still go through
  `core.execute("load_profile", …)` string commands with discarded
  results, rather than direct `KiwiCore` facade calls.
- Memoize the per-render `ProfileComposition.compose` in
  `SettingsModel.resolution(for:)`.

### Acceptance

- `editingProfile: String?` replaced by `EditTarget`; one reload
  path; no field reset duplicated across branches.
- Model-level tests cover live ⇄ stored transitions (dirty flag,
  savedSidecar cleared in stored mode, placementEditable, Lua
  editor mutually exclusive with stored editing).
- No behavior change for users; pure refactor.

### LINKED ISSUE #68: feat(gui): redesign Settings from scratch (simplicity-first capstone)
## Goal

Once **all** functionality that will be represented in the Settings
GUI has landed, redesign the GUI **from scratch** with a design
expert, optimising for **simplicity and intuitiveness**.

## Hard constraint

Every functional option that is settable in the GUI today must
**remain settable** after the redesign. What is open to change:

- **Where** each option lives (sections may be restructured,
  regrouped, renamed).
- **How** it looks (visual language, layout, interaction patterns).
- **What it's called** - goal is to make it easy to understand and intuitive
- Rarely-used / power-user settings may move into an **"Advanced"**
  tab or disclosure area to keep the common path clean.

What is **not** open to change: the set of behaviours the user can
configure. This is a re-presentation, not a feature cut.

## Process

1. **Brainstorm first** — information architecture and flows before
   pixels. (Can be driven with a `ui-designer` agent session:
   inventory every current control, cluster by task/frequency,
   propose a section structure + advanced split.)
2. Design pass with the design expert.
3. Rebuild the GUI against the agreed IA, preserving every setting.

## Scheduling (why it sits where it does)

- **Gated on Wave 7 landing** (app bars, per-space overrides,
  profiles UX, config cascade) — those waves add many new controls;
  redesigning before they exist means redesigning twice.
- **Must come before localization (Wave 8)** — a from-scratch UI
  changes most strings; localizing the old UI first means
  translating throwaway strings. Same "do it once" rule that keeps
  docs last: architectural → **redesign** → localize → document.

## Acceptance

- Full inventory of pre-redesign settable options exists and every
  one maps to a place in the new design.
- New IA reviewed and agreed before implementation.
- No configurable behaviour is lost.

_Roadmap capstone. Filed from a roadmap-restructuring pass._

---

### Must include: per-profile fallback-space chooser (from #75)

Expose a control to pick, **per profile**, the **fallback space** — the space that incoming windows land in when the user switches *to* this profile (spaces whose name the new profile doesn't define forward their windows there; see the profile-load space reconcile). Mechanism is already provided by #75 (move a chosen space to the first entry of the ordered space list); this is the UI to drive it. Drag-and-drop reordering of spaces (deferred from #75) is a natural companion to add here too.

### Must keep in mind
currently we can only controle enlarge and shrink window (no dimensionality. Later on it'll be possible to control it 2dim #56 . current 2 options turn to 4 (horizontal and vertical. Keep that in mind and mention in the issue after implementation on how to handle this in the GUI

### Must keep in mind — follow-ups from #71 that shape the design

- **Resize controls will grow.** #58 adds a global `resize.step`
  control (slider/stepper) next to Grow/Shrink; #56 later splits
  the 2 resize options into 4 (H/V) and may make the step per-axis.
  Reserve room in the resize area for both.
- **Animation duration becomes per-profile (#51).** It joins the
  five per-profile animation toggles; place duration with them and
  account for the `scroll.set_speed` overlap (same knob, two names).
- **Shortcuts tab UX requirements (Lane A: #33/#35/#34) —
  absorbed here.** Decision (2026-07-07): Lane A is *not* fixed
  beforehand — the bugs live in the recorder/`SettingsModel`
  wiring the rebuild replaces anyway, so they are acceptance
  criteria of this redesign instead: one active recorder at a
  time, conflict recompute on every mutation (including delete),
  inline duplicate feedback. Verify each against the new UI, then
  close #33/#35/#34 as folded-in.
- **Designated error surface (#39, #31).** The new design should
  include a visible place for config-load and profile-validation
  errors (e.g. menu-bar badge → details), so a half-loaded config
  is impossible to miss. (Only the *surface* is redesign scope —
  the typo-guard / validation cores stay with #39/#31.)
- **Adopt the `EditTarget` shape from #64 in the rebuild** (total
  enum + single reset-everything reload) instead of refactoring the
  old code first; close #64 as folded-in.

### Scope addition: menu-bar right-click menu (2026-07-07)

The redesign also covers the **status-item right-click menu** on
the menu-bar symbol, not just the Settings window. Inventory its
current entries the same way as the Settings controls, rethink
grouping/ordering with the design expert, and treat the two
surfaces as one coherent design (what belongs in the quick menu
vs. what belongs in Settings). Same hard constraint applies: no
action reachable today may be lost.

### Must keep in mind — branding & support

- **KiwiDesk logo (to be created later).** The IA/visual pass
  should designate where a logo lives once it exists — e.g. the
  menu-bar symbol itself, an About area, the Settings window
  header, and/or the onboarding walkthrough. Reserve the spot;
  the logo asset is its own later task.
- **Ko-fi donation link.** Add a small "Support KiwiDesk" /
  Ko-fi entry — candidate homes: the menu-bar right-click menu
  (now in scope, see above) and/or an About/footer area in
  Settings. Design pass decides placement; keep it discreet.

hajiboy95:
## Description

From-scratch rebuild of the Settings GUI and menu-bar quick menu per the approved IA (`plan/68-settings-redesign-ia-v2.md`), preserving every configurable behavior (re-presentation, not a feature cut).

**Navigation & chrome** — the flat TabView becomes a two-group sidebar that names the scope split: *This Profile* (Spaces, Layout Defaults, Monitors, Appearance, Behavior) vs *Whole App* (Profiles, Shortcuts, App Rules, General), with System-Settings-style icon tiles (§6.1). The footer collapses seven mode-dependent verbs into a stable Revert / Save a Copy As… / Save (⌘S) model with a target caption; Adopt moves into the raw-Lua editor's banner.

**Sections** — Spaces gains drag-reorder, the explicit fallback-space designation, mode glyphs, per-space icons, and inline per-row overrides (the second space picker is gone). Layout Defaults splits out the per-mode tuning; Monitors replaces canvas+palette+list with monitor cards holding space chips (pin/link micro-icons, context-menu keyboard fallback); Appearance gets uniform-first gaps with a teaching diagram, color wells + hex, a drag-visuals preview strip, and visible-but-inherited App Bar overrides with an active-override accent; App Rules unifies assignment+float lists into one row per app with GUI-assembled `App:Title` syntax; Shortcuts is regrouped by intent with a mode strip, header Import, and a collapsed Advanced Lua drawer; Profiles adds preset thumbnails and display-name monitor chips; General gains About (+ Ko-fi link).

**Menu bar** — header row naming the live profile, Switch Profile submenu, conditional Config Issues… entry backed by a new core config-issues surface (badge state: permission > config error > mode icon), Support link, entry icons (§6.2); standalone Config Issues window.

**Core** — explicit `fallback_space` reference (replaces first-of-order rehome; `KiwiDesk.set_fallback_space`), per-space icons (`space.icon`, `KiwiDesk.set_space_icon`), config-issues surface, a rename/delete fix for per-space maps with a reflection parity net, and reload-authority coverage for the new sparse settings.

**#64 folded in** — `EditTarget` enum + a single reload whose `TargetState` memberwise init makes forgetting a mode-dependent field a compile error; new `KiwiDeskGuiTests` target covers the live⇄stored transitions.

**Recorder acceptance (#33/#34/#35)** — one active recorder (synchronous teardown), duplicate hard-block with inline Steal/Go-to, conflicts derived live. Recording is **lock-on-full-release** (maintainer decision): the chord is snapshotted on key-down, live-previewed as it forms, and commits only when everything is released — staggered release order/timing can never corrupt the combo, and the base key can be corrected mid-chord (⌘J → ⌘K). Bare Escape, click-away, and app deactivation cancel; the state machine is table-driven-tested (ChordRecorderTests, 8 cases).

## Related Issue(s)

Closes #68, closes #64, closes #33, closes #34, closes #35

## Checklist

- [x] `swift build && swift test && ./scripts/lint.sh` passes
- [x] Release build passes: `swift build -c release`

## How I verified

- Full gate green at every commit: debug build, **634 tests / 118 suites** (incl. new FallbackSpaceTests, SpaceIconTests, ConfigIssuesTests, SpaceMapParityTests, EditTargetTests, DestinationParityTests, ChordRecorderTests), lint, release build.
- Two-agent review per AGENTS.md §3: parallel `code-reviewer` + `architect-reviewer` on the full diff (3 major + 15 minor findings — all fixed or consciously dismissed with reasons in the fix commits), then a sequential focused re-review of the fix range (**no major findings**), plus a third focused pass on the lock-on-release recorder commit (state machine verified sound; its two mediums — deactivation cancel, testable seam — fixed in the follow-up commit).
- Not yet exercised interactively in the running app — the checklist item above is checked for the automated verification; a manual pass over the new sections (drag-reorder, monitor-card drops, recorder flows) is recommended before merge. Happy to do a `/verify` run on request.

## Notes for Reviewer

- **Ko-fi URL is a placeholder** (`SupportLinks.koFi` → `ko-fi.com/kiwidesk`) — swap in the real handle before release.
- **#56 (2-axis resize) GUI slot**: reserved in Shortcuts ▸ Size & Float per §3.5/§5.5 — the single step control becomes a 2-row Horizontal/Vertical group above the (then four) resize rows; change is additive. A note is being posted on #56.
- **#58 (`resize.step`)** is not implemented here — only its slot; the catalog constant remains until #58 lands.
- Known cosmetic edge (dismissed in review): after a click-away recorder cancel, an immediate (<0.3s) re-click on the same Record button is swallowed once.
- Recorder manual-check note from review: local NSEvent monitors receiving keyUp while ⌘ is held is how established recorder libraries work, but the ⌘-held-release-base-last path deserves one pass in the running app.
- The IA plan's guessed `layout.scroll.speed` key shipped in #51 as `animations.scroll_speed`; the GUI binds to what shipped.

https://claude.ai/code/session_01DqnfUBrxAMCxUTrU1SGuXx


hajiboy95:
### Update — settings-polish pass (fd2a82e, 57729cc, 2d394d5)

Added on top of the existing branch:

- **fd2a82e** — settings redesign polish: stacked full-width header
  (title + profile dropdown + conditional status), full-height sidebar,
  Design/System sidebar groups, App Bar overrides behind a disclosure
  (lag fix), lightweight color swatch, editable steppers, General
  reorder, dark-mode wordmark + runtime Dock icon + `@MainActor`
  `BrandAssets`.
- **57729cc** — review fixes: `ColorPanelController` resign-token
  lifecycle, guarded `.accessory` demotion, `StepperRow` `!focused`
  guard, `showsProfileContext` parity test.
- **2d394d5** — Config Issues window releases `.regular` on close
  (last activation-policy edge).

**Review:** `code-reviewer` + `architect-reviewer` ran in parallel,
then a sequential re-review of the fix range; all findings addressed
(no open items).

**Verify gate:** `swift build` ✓ · `swift test` (659) ✓ ·
`scripts/lint.sh` ✓ · release build ✓ (strict-concurrency clean).

Still pending (separate follow-ups): the golden dark-mode logos
(wordmark + app mark, palette drafting) with PNG regen, plus #89
(`.app` bundle/`.icns`) and #90 (sidebar search).

Issue #406, opened 2026-07-19
Full-app UX audit pass: labels, controls, discoverability vs north-star (pre-launch rename window)


hajiboy95:
## Why now

Settings has grown fast — bars, profiles, layouts, overrides, palettes, borders, onboarding — and vocabulary/control choices were made incrementally. The recent bars color-model merge (`box`/`active-box`/`background` → **Fill**, `text` → **Item**, `hover` → **Hover fill/item**, front-app/focused-window unification) showed how much drift accumulates before anyone looks across the whole surface.

We are **pre-launch, single-user → zero migration cost** (AGENTS.md §5): renames are free now and expensive forever after. Spend that leverage once, deliberately, before launch.

## Ask — three-agent pass

- **`ui-ux-tester` (first)** — exhaustive functional walk of every Settings flow: each control **applies**, **persists** across save/reload, and is **reachable**; report defects, dead ends, and unreachable/inert controls, structured. Establishes what actually works before design judges it.
- **`ui-designer` (second)** — judges the now-known-working surface against the three north-star priorities **in order**: (1) simplicity, (2) intuitiveness, (3) Apple-native feel — plus *approachable by default, powerful on demand*.
- **`accessibility-tester` (parallel with designer)** — VoiceOver label coverage (every `L()` string → an `accessibilityLabel`), full keyboard navigation/focus order, contrast of the shipped palette defaults, Dynamic Type, reduce-motion. **Priority focus:** color-independent state cues — after the bars color-model merge, active / hover / focused states must not be distinguishable by **color alone** (a shape/indicator cue must accompany every color cue). Flag any color-only state.

## Scope — every user-facing surface

- Every **label**: setting names, section titles, captions, option strings — consistent vocabulary, no synonyms/drift, says-what-it-does.
- Every **control choice** (toggle / segmented / dropdown / slider) and **row tier** (inline vs Advanced disclosure) per `docs/ui-patterns.md`.
- **Discoverability** — can a newcomer *find* each setting? Sidebar structure, search, contextual `?` help, grey-don't-hide (#171).
- **Grouping** — group-by-topic; toggle sits directly above the control it gates.
- **Cross-bar / cross-tab consistency** — same concept → same word everywhere.

## Deliverable

[...]

Issue #277, opened 2026-07-15
💡 Fill the Settings control catalog so search lands on the control


hajiboy95:
### Use case

Settings search can find a **destination** and a **section**, and
it can land you on any control the catalog already carries — but
the catalog carries a fraction of the controls that exist.
`SettingsSearchIndex` says so at its own declaration, and
`SettingsControl` names the gap it leaves: a drawer whose
interior is not cataloged has no anchor, so a hit inside it falls
back to the enclosing section and the user finishes the search by
eye.

That is the whole of what is left here. The three *mechanisms*
this issue was opened for have all shipped:

- **the control catalog itself** — `SettingsCatalog` /
  `SettingsControl` / `SettingsSearchIndex`, with the
  parity-test obligation met (`SettingsCatalogTests`,
  `SettingsCatalogArgumentTests`, `SettingsCatalogSiteTests`,
  `SettingsAnchorPrimitiveTests`);
- **scroll-to + highlight** — `SettingsAnchor`,
  `SettingsView+Reveal` and `SettingsReveal`, whose transient
  accent wash was settled by a `ui-designer` consult rather than
  the ring or pulse first sketched;
- **layout-mode auto-select** — the mode selection lifted off
  the section's view-local `@State` onto `SettingsNavigation`,
  so a hit inside a per-mode editor selects that mode on
  landing (`SettingsSearchAnchorTests`).

So this issue is no longer "tier 2". It is the fill, and it is
retitled to say so. Its old title also named a **sidebar** that
no longer exists — #678 turn 9 replaced it with Home plus the
header search field.

### Proposed solution

Grow the catalog until a search hit lands on the control itself
rather than its section — but **not for every row, and the
difference is measurable** (audited 2026-08-27 against the live
`SettingsSearchIndex`).

**165 census rows carry no anchor, and they are not one
population.** Split by `SettingPlacement.tier`:

| Tier | Rows | What an anchor buys |
|---|---|---|
| `atRest` | **96** | A highlight wash on a control that is **already on screen** when you land on its section |
| `showMore` | **70** | **The drawer opens.** Without an anchor the hit falls back to the section and the control stays hidden behind a shut disclosure |
| `immediate` | 1 | — |

That asymmetry is the whole scope question, and it follows from
a mechanism already documented at `SettingsAnchor.swift`: a
collapsed `DisclosureGroup` still renders its own header, and
`SettingsDisclosure` expands when the standing reveal target is
[...]

PR #514, opened 2026-07-25
fix(gui): make the Advanced drawers findable in sidebar search (#277, #406)

hajiboy95:
Refs #277, #406. **GUI — held open for your eye-confirm; it changes what the sidebar search returns.**

This is the **tier-1 completeness** half of #277, found by the #406 audit. Tier 2 proper (per-control catalog, scroll-to + highlight, mode auto-select) is **not** in here — see the bottom.

## The gap

Typing **"advanced"**, **"init.lua"** or **"fingerprint"** into the sidebar search returned nothing. Three real, static drawer headers were never in the index:

- General's **"Advanced"** — which holds the config-file path and **"Edit init.lua directly"**, arguably the most important escape hatch in the app
- Shortcuts' **"Advanced: Lua bindings"**
- Monitors' **"Advanced — monitor fingerprints"**

Also missing, same class: Bars' **"Advanced colors"** and Appearance ▸ Gaps' **"Per-edge…"** / **"Per-axis…"**.

A drawer is precisely what a user searches for. It's collapsed, so its contents are exactly the ones they can't find by scanning the pane — which makes "invisible to search" the worst possible place for the gap to be.

## Why no test caught it

`SidebarSearchParityTests`' reverse guard ("every rendered header is indexed") matched only `SettingsSection(L(…))`. Every one of these is a `DisclosureGroup`, so they were **structurally invisible to the scan** — not an oversight in the list, a blind spot in the lens. The guard now also requires every drawer title to be indexed or consciously excluded.

That half is **hand-listed, and the comment says why.** I tried the scan first, and it doesn't work here: SwiftUI's label may lead the initializer (`DisclosureGroup(L(…), isExpanded:)`), trail the body (`… { body } label: { … }`), or live in a separate computed property far from the call — so no anchor reliably picks the title out, and in practice it kept selecting a row from *inside* the drawer. A guard that needs a growing exclusion list to stay green is worse than an honest short list. The trade-off (the list is itself a place to forget) is written into the comment rather than glossed.

Verified non-vacuous: adding an unindexed drawer key turns it red; removing it turns it green again.

## Deliberately not indexed

Search navigates to **destinations**, and these two have none to offer:

- `space_override.dormant.label` — inside the per-Space "Overrides…" popover, reachable only from a row button.
- `app_bar.layout.overrides` — inside the per-layout section whose computed title (`"%1$@ bar"`) is already the deferred case in this file; indexing the child while the parent is deferred would point search at a header the result caption can't name.

## What tier 2 still needs

Not started, and it needs plumbing that doesn't exist yet — `ScrollViewReader` anchors on every section and row, plus lifting the Layout Defaults mode selection out of view-local `@State`. Two things are now in place for whoever picks it up:

- **`plan/settings-control-inventory.md`** — an exhaustive per-control catalogue produced during the #406 audit: label key, owning destination, sub-group and visibility rule for every control. That is the catalogue tier 2 has to build, already gathered.
- Its §B lists the three **non-`DisclosureGroup` drawers** the scroll-to plumbing must also reach (the Overrides popover, the Bars App Bar/Space Bar switch, the Layout Defaults mode strip), and §C the one-key-many-sites controls the catalogue will have to disambiguate.

## Verify gate

`swift build` · `swift test --skip ExecTests` (**1931**) · `swift test --filter ExecTests` (**14**) · `scripts/lint.sh` (exit 0) · `swift build -c release` — all green.

## Eye-confirm

Type "advanced", "init.lua", "per-edge" and "fingerprint" into the sidebar search and check each lands on the right tab. No pixels change; only the result list does.


hajiboy95:
## Review round 1 — two blocking findings, both valid, both fixed

Posting because the first cut of this PR went up **before** its review ran (my process error). `code-reviewer` then found that the guard I'd added was **fail-open for the exact regression it was written for**, and that my stated reason for hand-listing was wrong. Both correct.

### 1. The guard didn't guard

`renderedHeadersAreIndexed` iterates `headers` = regex-matched `SettingsSection` keys + the hand-listed `drawerTitles`. A **new** drawer is in neither, so it was never examined — a new Advanced drawer could ship unsearchable and no test could fail. That is the #277 bug repeating verbatim. The hand list only bought the reverse direction (an existing drawer's key can't be dropped from the index), which is not what my comment claimed.

### 2. A reliable scan does exist — I gave up too early

I claimed no anchor could find a `DisclosureGroup`'s label. I'd only tried flat regexes. The trap is real (`GeneralSection`'s drawer body contains `Button { … } label: { Label(L("…edit_lua", …))`, so "next `label:`" picks a control *inside* the drawer) — but walking the actual `(…)`/`{…}` nesting resolves **9/9 sites → the same 8 keys**, zero exclusions.

`disclosureTitleKeys(in:)` now does that, skipping string literals and handling both real shapes. A site it can't resolve is returned **unresolved and fails loudly** rather than skipped. **Proven fail-shut:** adding a real `DisclosureGroup` to `BehaviorSection` without touching any test turns it red; removing it turns it green.

### 3. The same blind spot had a second victim I'd missed

`drag.ghost` / `drag.drop_zone` — `DragVisualsEditor.column(title:)` passes a **variable** first argument, so those headers are invisible to the scan by construction, weren't indexed, and weren't excluded. Searching "Ghost" or "Drop zone" returned nothing, and nothing could ever go red. Now indexed.

### Also fixed
- `bars.advanced_colors` was appended last, breaking `subsections(of:)`'s render-order contract (it renders inside App Bar Global colors, and the first match wins the caption).
- The `space_override.dormant.label` exclusion now records its durable reason: the string carries a runtime count, and the index stores formatted strings — so it can't be indexed without inventing an argument.
- Two **behaviour** tests added. Both parity tests compare keys, so neither can see the destination mapping — the blind spot this file's own header comment names.

### Correction to the PR body above

**"init.lua" still returns nothing.** `general.advanced.edit_lua` is a *control* label, i.e. tier-2 territory. What this PR fixes is that the drawer holding it is now reachable by searching **"Advanced"**. The original description overstated it — please read the first section of this PR with that correction.

### Gate after the fixes
`swift build` · `swift test --skip ExecTests` (**1933**) · `swift test --filter ExecTests` (**14**) · `scripts/lint.sh` (exit 0) · `swift build -c release` — all green.

Issue #196, opened 2026-07-12
feat(app): block second KiwiDesk instance (single-instance lock)


hajiboy95:
Part of #195 (app-lifecycle collector).

## Problem

Multiple KiwiDesk instances can launch at once. Two running
instances both drive AX/SkyLight, register the same Carbon hotkeys,
and fight over window placement — undefined, racy behavior. Nothing
today stops a second launch.

## Expected

- If KiwiDesk is already running, a second launch does not start a
  second instance.
- Instead: focus/surface the existing instance (or no-op with a
  brief notice), then exit the new process.

## Approach (proposal)

1. `NSRunningApplication` check by bundle ID at launch: another
   instance of our bundle ID running → activate it and `exit(0)`.
   AppKit-native, covers the `.app` path.
2. `flock` on a file in Application Support as a belt-and-suspenders
   lock, so dev/CLI launches (no proper bundle) are covered too.
   Auto-releases on process death (no stale PID file to
   hand-validate).

Check the IPC layer (`Sources/KiwiDeskCore/IPC`) first — it may
already hold a socket/port we can reuse as the liveness signal
instead of a new lock file.

## Notes / edge cases

- Stale lock after a crash must not permanently block relaunch —
  `flock` auto-releases, so relaunch stays clean.
- Wedged/hung running instance: default = surface an error, don't
  force-kill.
- CI/tests must not trip the lock — gate behind the real app
  bootstrap, not `KiwiDeskCore`.

## Docs

- `docs/user-guide.md`: note single-instance behavior.

Issue #96, opened 2026-07-09
i18n: localize Core-layer UI strings (conflict/system-shortcut tooltips)


hajiboy95:
### Context

Follow-up from #9 (localization). The GUI is now routed through `L()`, but a set of user-facing strings that originate in **`KiwiDeskCore`** remain English because they can't currently use the lookup.

### The gap

- `Sources/KiwiDeskCore/Keys/KeybindingConflicts.swift` — e.g. `"Not a recognized shortcut."`
- `Sources/KiwiDeskCore/Keys/SystemShortcuts.swift` — ~14 hardcoded macOS shortcut descriptions ("Spotlight", "Close Window", …) shown as conflict tooltips.

These reach the UI as tooltips but are never translated.

### Why they were left out of the #9 sweep

`L()` / `LocalizationManager` are `@MainActor`-isolated (the manager drives SwiftUI and publishes). The two files above are intentionally **actor-free, pure Core logic** exercised by non-`@MainActor` unit tests (AGENTS.md §2.6). Routing them through `L()` as-is would force MainActor isolation onto code designed to stay actor-free and testable — a materially larger architectural change than a string wrap.

### Options to evaluate

1. Resolve these strings to a localized form **at the GUI boundary** (where a `@MainActor` context exists) rather than inside Core — Core returns a stable key/enum, the GUI localizes it. Mirrors the keybinding-label `canonical → resolvedLabel` split already used in `KeybindingCatalog`/`StandardLayoutDisplay` (see commit cd58af2).
2. A non-`@MainActor` lookup path for these specific pure strings (needs care re: the manager's isolation).

Option 1 is likely the cleaner fit and keeps Core actor-free.

Follow-up backlog: #71 · relates to #9

PR #574, opened 2026-07-28
feat(service): open at login via SMAppService (#342)

### LINKED ISSUE #342: feat(service): use SMAppService for the login item instead of a hand-written LaunchAgent plist
## Context

`service start/stop/restart` writes
`~/Library/LaunchAgents/org.kiwidesk.KiwiDesk.plist` and drives it
with `launchctl bootstrap`/`bootout` (`ServiceManager.swift`).

**This issue was originally justified by the App Store**, whose
sandbox forbids both writing that path and calling `launchctl`.
That framing is dead: the App Store is out of scope, settled as a
`[Principle]` in `docs/design-decisions.md` ("Distribution: direct
download, not the Mac App Store").

Per the discussion on #133, the issue is **rewritten, not closed** —
`SMAppService` is the modern login-item API for directly-distributed
apps too, so the idea outlives its original motivation. What changes
is the argument: this is now a **UX and robustness** improvement, not
a compliance requirement, and nothing blocks on it.

## Direction

`SMAppService.mainApp` registers the app as a login item the user
toggles in System Settings ▸ General ▸ Login Items. The app *is* the
service; registering it makes it auto-launch at login. The user-facing
shape is an **"Open at Login"** toggle in Settings rather than CLI
verbs.

Why it is still worth doing without the App Store forcing it:

- A hand-written plist plus `launchctl` is fragile and invisible —
  the user cannot see or revoke it where macOS trains them to look.
- macOS surfaces `SMAppService` registrations in System Settings, so
  the app stops being a thing that silently installed a LaunchAgent.
- It removes bespoke plist authoring from `ServiceManager`.

## Scope / open questions

- **Do both mechanisms coexist?** `service *` verbs are useful for the
  CLI and for scripted setups, so the likely answer is the toggle and
  the verbs drive the same underlying registration rather than two
  competing ones. Needs confirming: if the service is already loaded
  via `launchctl`, does the toggle read as on? Two sources of truth
  here is the failure mode to avoid — the same "one owner of updates"
  reasoning as the cask's `auto_updates` trap.
- Whether the toggle belongs in the onboarding flow after the
  Accessibility grant — ask a `ui-designer` consult.
- Where the toggle lives in Settings (General?).
- What, if anything, `service *` verbs report once the login item is
  the primary mechanism.

Depends on #89's real `.app` bundle (**done** — `febc5264`), so this
is unblocked. Not string-frozen yet.

hajiboy95:
Implements #342 — a user-facing **Open at Login** control, using `SMAppService.mainApp` (the modern macOS 13+ login-item API) rather than a hand-written LaunchAgent. Resolves the issue's open onboarding/placement question per the ui-designer consult posted there.

## What it does
- **`LoginItemManager`** (Core, `Service/`): wraps `SMAppService.mainApp` register/unregister/status; maps OS status → a 4-case `LoginItemState` (`notRegistered` / `enabled` / `requiresApproval` / `unavailable`). Core returns structure, the GUI renders the sentence (#96).
- **Settings toggle** (`Settings/LoginItemCard.swift`), top of **General**: **read-through** — reads live `SMAppService` status on appear + `didBecomeActive`, never caches a bool, so a change made in System Settings ▸ Login Items directly is reflected. `.requiresApproval` reads *on* with an "Open Login Items" jump; `.unavailable` greys the control (grey, don't hide).
- **Onboarding**: a pre-checked checkbox on the closing card; both exit routes commit the choice. Default on because a tiling WM's off-state after reboot is a desktop of unmanaged windows, not a neutral absence.
- Deliberately **separate** from the `kiwidesk service` LaunchAgent + `KeepAlive` (crash-restart supervision), not reconciled here. Documented overlap: that agent's `RunAtLoad` also auto-starts at login, invisibly to the toggle — accepted, made safe by the #196 instance lock.

## Design direction
Recorded as a `[Principle]` in `docs/design-decisions.md` (pre-checked default; **no** per-launch modal — same standing-nag shape already rejected; read-through toggle). User-guide documents the control.

## Tests
- `LoginItemStateTests` (Core): pure status mapping + `isOn` derivation.
- Onboarding: default pre-checked; both exit routes commit the choice.
- Catalog guard counts bumped (49 declarations / 55 entries).

## Verification
`swift build` clean · **2110 + 14 tests pass** · lint OK · `extract-keys --check` OK.

## Review
Round-1 `code-reviewer` + `architect-reviewer` (parallel): **no correctness bugs**. Findings were doc/comment-honesty only (the "single authority" overstatement + a false "CLI verb reflects through the toggle" claim from the `RunAtLoad` overlap) plus a trivial `@State` probe cleanup — all addressed in the second commit.

fixes #342


hajiboy95:
## Device QA fix + i18n + General polish (commit `cda09a53`)

Device testing surfaced a real bug and some polish, all addressed:

**Bug — the toggle greyed itself out.** macOS reports `SMAppService.mainApp.status` as `.notFound` *before* first registration (verified on 26.5), and the code folded `.notFound` → `.unavailable` (greyed), blocking the tap that would register it. Confirmed against device QA: registering from /Applications works and the item shows in System Settings ▸ Login Items. Fix: `.notFound` → `.notRegistered`. The *terminal* `.notFound` (a bare non-bundled binary, or a Gatekeeper-translocated download — where registration genuinely can't succeed) is separated by a location check in `current` and greys out with an actionable **"Move KiwiDesk to your Applications folder to turn this on."** caption.

**GUI polish (2nd ui-designer consult):**
- General order → **Language, then Login** (Language is the only place that decision is made; Login is already set in onboarding).
- Split the doubled label the way every other card does: heading **"Login"** + toggle **"Open KiwiDesk at Login"**; Language keeps its heading, its picker becomes **"Display language"**.

**Translations:** all 8 new/changed keys translated across **all 11 locales** (login-item strings, onboarding checkbox, "Display language"), using each platform's own Login Items / System Settings / Applications terminology. Localization content guards (residue, overlap, product-name) pass.

Gate: build clean · 2110 + 14 tests · lint + `extract-keys --check` OK. Verified working on device (toggle registers, appears in Login Items).

Issue #606, opened 2026-07-28
💡 Export / import your setup (gui.json + profiles) to move it between Macs


hajiboy95:
> No milestone yet — a later feature. Captured from a `ui-designer` consult as a discussion-starter; the open decisions below need owner sign-off before scoping.

## Use case

Let a user **export their KiwiDesk setup and import it on another Mac** — the one-time "new machine" migration and pre-change backup case. The bundle carries the GUI-managed config (`gui.json`) and all their profiles, so a setup moves in one step instead of hand-copying files.

## Scope (recommended)

Build a **light version, scoped to migration/backup, GUI-managed only.** The three motivations don't share a solution:

- **Continuous multi-Mac parity** — *not* this feature's job (a snapshot goes stale immediately). Already solved free by symlinking `~/.config/KiwiDesk` into iCloud Drive/Dropbox. Worth a docs callout, not code. Consider to place it maybe in the general tab or somewhere? needs second thought.
- **One-time migration / backup** — **yes, build it.** Curating "which files travel" is exactly the job the GUI should own for a non-technical user.
- **Sharing one setup with someone** — a *single-profile* need, nearly free given `ProfileManager` is already one JSON per profile with a reveal path. Track as a separate, smaller "Export/Import profile…" ask (see open decision 2).

## Placement

**General ▸ Advanced**, beside the existing "Configuration file / Reveal" row (`GeneralSection`) — not a dedicated tab. Matches the Apple-native precedent (Safari "Manage Website Data", Mail mailbox export live in an existing pane's advanced corner) and extends KiwiDesk's existing "point Finder at the real file" instinct rather than replacing it.

## Bundle contents

| Include | Exclude | Why |
|---|---|---|
| `gui.json` | `init.lua` | Lua-managed users own their portability by choosing that path; folding it in implies KiwiDesk manages a file it deliberately doesn't. **Hide the whole feature when `!isGuiManaged`** (removed, not greyed — the "channel doesn't exist for you" case, App Store-badge precedent). |
| every file in `profiles/` | `UserDefaults` language pref | Machine/account-scoped, not config-scoped; silently changing a new Mac's language on import is an unasked-for side effect. |
| (palettes ride inside `gui.json` — verify) | separate palette resource? | Appearance is "one-shot apply" — colors likely written *into* `gui.json` at apply time, so they travel for free. **Verify against `Appearance` code** (open decision 1). |

## Import semantics

[...]

Issue #602, opened 2026-07-28
💡 Default ⌃⌥K shortcut to open the Shortcuts panel (+ auto-seed per mode)


hajiboy95:
## Use case

KiwiDesk ships a read-only Shortcuts cheat-sheet panel (#326, `KiwiDesk.show_shortcuts()`), but **no default keyboard shortcut opens it**. Verified in `DefaultKeybindings.swift`: the first-run seed covers focus/move/swap/resize/go-to-space/toggles, but not `show_shortcuts()`. So on a fresh install the *only* way to open the cheat-sheet is mouse → menu-bar "View shortcuts". For a keyboard-first tool, its own discoverability surface is unreachable from the keyboard by default.

Compounding it: keybindings are **strictly per-mode** (there is no global binding layer — `KeybindingManager.activate(currentMode)` swaps the active mode's bindings in/out). A newly-created mode starts with **zero** bindings (`ShortcutsHeader.addMode()` → `KeyMode(name:)`), so even if we seed a default into the base mode, a custom "open applications" mode would have no way to open the panel from the keyboard.

## Proposed solution

Seed `KiwiDesk.show_shortcuts()` bound to **⌃⌥K** (control+option+k) as a normal per-mode row:

1. **Base default** — add the row to the first-run seed in `DefaultKeybindings.swift` (via the `KiwiCore+GuiConfigSeed.swift:221` seed path), same shape as the existing ⌃⌥f / ⌃⌥s toggle rows.
2. **Every new mode** — auto-seed the same ⌃⌥K row when a mode is created (`ShortcutsHeader.addMode()`), so a fresh custom mode can open the panel immediately.

It's an **ordinary editable/deletable row per mode** (Option A, not a true global binding) — a user can rebind or clear it in Settings ▸ Shortcuts per mode like any other default; deleting it in one mode doesn't touch others. The menu-bar "View shortcuts" item stays the un-loseable, mode-independent fallback.

### Why ⌃⌥K (design consult with `ui-designer`)

[...]

Issue #596, opened 2026-07-28
Focus ring settle-tail: echo hiccup, stranded-ring heal timing, mid-flight sync rewind


hajiboy95:
Follow-up cluster from the #594 reviews (2 LOW + 1 MEDIUM, deferred
pending device evidence). All three live at the *temporal edge* of
the #594 guard — the moment the animation clock settles while a
slow-AX app is still applying queued frames — and share one likely
cure, so they are filed together.

## 1. Post-settle backward hiccup (AX-fallback path only)

`FollowSource.applies` gates echoes on `isAnimating`, whose edge is
"the animation *clock* settled" — but on slow-AX apps the app is
still applying queued frames for 100–300 ms after that instant. The
first post-settle `.windowMoved` echo (fresh AX read, but the app is
still behind) passes the guard and pulls the ring *backward* off the
target the settle tick just applied, then later echoes walk it
forward again. Only visible when `skyLightActive` is false (WS
stream down); pre-#594 the ring trailed the whole flight, so the
tail was invisible. #47's per-tick grow (`growRateHz = nil`) makes
slow-AX apps maximally behind at settle, so the tail is largest for
grows.

Not fixed with a blanket `didRecentlySetFrame` reuse: the instant
`setFrame` path deliberately does not tee, so under AX fallback the
echo is the only carrier of ring updates after an instant set — a
1 s echo suppression would freeze the ring there. Needs a narrower,
animation-scoped grace (stamp per-window off
`onAnimationEnd`/`onWindowSettled`, OR into the managers'
`isAnimating` closures — wiring-only in `KiwiCore+Bootstrap` thanks
to the injected-closure seam).

## 2. Fully-unresponsive app strands the ring; the heal fires too early

If an app accepts *no* AX writes for the whole flight (beachballed),
real bounds never change → zero WS events, zero echoes — so the
"first WS event after settle reconciles residual drift" never
happens and the ring parks at the commanded target over an unmoved
window. The intended heal, `scheduleBorderDropReconcile`
(`KiwiCore+Borders.swift`), fires at `durationMS + 50` — but the
spring's visual settle is ~2× its response (`response =
durationMS/1000 × 1.4`), so for the default 250 ms it lands
mid-flight and the remaining ticks re-yank the ring; nothing
corrects after settle. `animationsDidSettle` runs no border pass.

Fix direction: drive the drop-reconcile (or a border re-sync) from
[...]

PR #608, opened 2026-07-28
feat(onboarding): ⌃⌥K hint on the ready card + config-sync docs

### LINKED ISSUE #604: 💡 Onboarding: point new users at ⌃⌥K for the shortcuts cheat-sheet
## Use case

A new user gets a working ⌃⌥-based tiling setup on first run, and (via #602) ⌃⌥K now opens the read-only Shortcuts cheat-sheet in every mode. But nothing in onboarding points them at ⌃⌥K, so the discoverability surface stays undiscovered until they happen on the menu-bar "View Shortcuts…" item. KiwiDesk is keyboard-first — "here's how to see what you can press" is core-value teaching, not a feature tour.

## Proposed solution

Fold **one clause** into the existing "You're ready to go" onboarding card (`readyToExplore` in `OnboardingView.swift`) — no new step, no new card, same one-time `OnboardingDiscovery` gate. Extend the closing sentence that already teaches "how to get back here":

- Current: *"…You can always get back here from the menu bar."*
- New: *"…You can always get back here from the menu bar — or press ⌃⌥K for a shortcuts cheat-sheet."*

**Do not** also say the shortcuts are adjustable: the panel itself carries that hand-off (its "Edit in Settings…" footer, the #326 glance→edit bridge). Teaching editability in onboarding, before the user has seen the panel once, front-loads config that approachable-by-default says to avoid. Onboarding's job here is just "press ⌃⌥K".

## Alternatives considered

- A dedicated onboarding step/card for shortcuts — rejected as feature-tour bloat; one clause on the existing closing card is the lightest form.
- Mentioning "and you can rebind them" — rejected (see above).

## Extra context

Design consult (`ui-designer`) framed this against the north-star (simplicity, intuitiveness, Apple-native; approachable by default): Apple avoids keystroke-teaching in onboarding because those apps aren't keyboard-first — KiwiDesk is, so this one line earns its place, once.

**Kept out of the #602/#603 branch deliberately.** #602 (seed ⌃⌥K) and #603 (auto-close on mode switch) are mechanical keybinding/controller changes with no localization. This edits a **shipped localization key's meaning**, so per the l10n rules it needs `scripts/drop-key` + re-translation across all 11 locales — real, separate tail work that shouldn't ride invisibly on a keybinding PR (one-branch-one-change).

Depends on #602 landing (so ⌃⌥K is the stable, correct combo to cite).

hajiboy95:
Two doc/copy items from the shortcuts-panel work.

## #604 — onboarding points at ⌃⌥K
The first-run "You're ready to go" card never mentioned that ⌃⌥K opens the read-only shortcuts cheat-sheet (seeded in #602). One clause added to its closing sentence — no new step/card, same one-time gate. Deliberately omits "and they're editable" (the panel's "Edit in Settings…" footer carries that; front-loading config fights approachable-by-default).

English meaning changed → `onboarding.ready.body` dropped and **re-translated across all 10 locales** (`drop-key` + `merge-keys`; content guards clean, ⌃⌥K/KiwiDesk kept verbatim, cheat-sheet localized per language).

## Config-sync docs (refs #606)
A low-key power-user tip in `docs/user-guide.md`'s gui.json section: symlink `~/.config/KiwiDesk` into iCloud Drive/Dropbox for continuous cross-Mac sync — the *live-sync* answer, distinct from the future one-time export/import (#606). States what doesn't travel (Accessibility grant, display layout, Spaces). Flows to the site via the existing docs symlink (no new page/nav → no site build).

## Verification
- `swift build` ✓, `swift test` ✓ (2140), `scripts/lint.sh` ✓ (`extract-keys --check` OK, localization guards clean).
- Copy + placement per `ui-designer` consult. No logic/architecture change — two-agent review not triggered (§3.6).

fixes #604


hajiboy95:
Landed on `main` via fast-forward as 418ece3f + 87589011 (rebased onto #596's c6e3e9d6, which merged concurrently — hence the SHA change GitHub couldn't auto-match). Closing manually.

PR #665, opened 2026-08-02
fix(localization): make "System default" follow the system language

### LINKED ISSUE #659: 🐛 System default language always resolves to English in the packaged app

0.9.4

### macOS version

macOS 26.6

### Mac model type

### Steps to reproduce

1. Set the macOS system language to a non-English language that
   ships a catalog (System Settings → General → Language & Region;
   e.g. Deutsch).
2. Package the app: `./scripts/build-app.sh`.
3. Launch the `.app` and open Settings → General → Language,
   leaving the picker on **System default**.

**Expected:** the UI renders in the system language whenever a
catalog for it ships — there are 11 in
`Sources/KiwiDeskCore/Resources/Locales/`.

**Actual:** the UI is English. Picking the language explicitly in
the same picker works, so the catalogs themselves are fine; only
"System default" is broken.

Two independent causes, either of which alone is enough to produce
the symptom:

1. **`scripts/build-app.sh` writes no `CFBundleLocalizations`.**
   The generated `Info.plist` declares `CFBundleDevelopmentRegion`
   (`en`) and nothing else, so macOS resolves the process locale to
   the development region no matter what the user's language list
   says — the bundle claims to speak one language, so that is the
   one the OS hands back. `LocalizationManager.effectiveLocale`'s
   system branch therefore could not return anything but English.
   This does **not** reproduce under `swift run`, which has no
   `Info.plist` at all, which is how it survived this long.

2. **The match was on the bare language subtag.**
   `effectiveLocale` compared
   `Locale.current.language.languageCode` against the shipped
   list. That collapses `zh-Hant` and `zh-Hans` to `zh`, and
   `pt-BR` to `pt` — none of which name a shipped catalog. So
   Chinese and Brazilian-Portuguese systems would still have fallen
   through to English even after cause 1 was fixed.

### Relevant output and logs

On a German Mac, `Locale.preferredLanguages` is
`["de-DE", "ja-DE", "en-DE"]` and a plain Swift script resolves
`Locale.current` to `de_DE` — while the identical lookup inside the
packaged `.app` resolves to English. That gap between the two build
shapes is the whole bug.

- [X] None that I know of

### Extra context

Fix in progress on `fix/system-default-language`:

- `build-app.sh` derives `CFBundleLocalizations` by globbing the
  shipped catalog directory, never a typed list — `en.json` is on
  disk as the translator manifest, so English joins by
  construction rather than by a special case.
- A new pure `LocaleMatch` helper replaces the bare-subtag
  comparison: it walks `Locale.preferredLanguages` in the user's
  own priority order, tries each tag exactly and then through
  `maximalIdentifier` (so `zh-TW` becomes `zh-Hant-TW` before it is
  ever shortened to `zh`), widens a bare language onto a refining
  catalog only as a last resort (`pt` → `pt-BR`), and
  short-circuits English wherever it sits in the list — without
  that last part, an English-first user silently gets their
  *second* language.

hajiboy95:
Fixes #659.

Two independent bugs, either alone enough to force English on every non-English Mac.

**The packaged `Info.plist` declared no `CFBundleLocalizations`.** A bundle that names one localization gets that one back: macOS resolved the process locale to `CFBundleDevelopmentRegion` whatever the user's language list said, so `effectiveLocale`'s system branch could not return anything else. It never reproduced under `swift run`, which has no `Info.plist` at all — which is how it survived.

**And the match was on the bare language subtag**, which collapses `zh-Hant`/`zh-Hans` to `zh` and `pt-BR` to `pt`, none of which name a shipped catalog. Chinese and Portuguese systems would still have fallen through once the plist was fixed.

`build-app.sh` now derives `CFBundleLocalizations` by globbing the shipped catalog directory rather than carrying a typed list — `en.json` is on disk as the translator manifest, so English joins by construction. `LocaleMatch` replaces the subtag comparison: it walks `Locale.preferredLanguages` in the user's own priority order, resolves each tag exactly over its `maximalIdentifier` truncations before widening a bare language onto a refining catalog, and short-circuits English wherever it sits.

The phase ordering is the subtlety and has its own test: widening per truncation step instead of after all of them hands `zh-TW` the **Simplified** catalog, which is the wrong language rather than a coarser fallback.

## Guards, and what they cost to get right

`guard-prover` broke the plist guard twice before it held:

- emptying the `<array>` passed every check — shipping a plist declaring *zero* localizations, which is this bug verbatim;
- assigning `LOCALE_KEYS` a hand-typed list *after* the glob loop passed too, because the checks read three disconnected places.

It now asserts the array is non-empty *and* carries the expansion, and follows `LOCALE_KEYS` end to end. Review then found a third: quoting the heredoc delimiter (`<<'PLISTEOF'`) ships the literal `$LOCALE_KEYS` while everything still passed. Asserted and red-proved.

`LocaleMatchTests` had the same shape of hole — a bare `zh` maximalizes to `zh-Hans-CN` and exact-hits, so it never reached `widen`, and with one `pt-*` catalog shipped the `.sorted()` determinism was untested. There is now a case with two Portuguese catalogs passed in both input orders.

**One claim was walked back.** The docstring said both halves of the fix were necessary; that was never measured. `Locale.preferredLanguages` is documented as the *user's* list rather than the bundle-resolved one, so the matcher change alone may be what fixes this path. The claim is now scoped to what was observed.

## Verification

`swift build` · full `swift test` (2328 passed) · `scripts/lint.sh` — all green.

## Review

`code-reviewer` and `architect-reviewer`, all findings addressed. `architect-reviewer` confirmed `LocaleMatch` earns being its own type and that the plist/matcher split is the right seam.


hajiboy95:
Device QA done (2026-08-02, macOS 26.6, M1 Max, signed builds of both sides).

**Both halves of the A/B ran on one machine, minutes apart**, with the picker on *System default* (the persisted `language = de` was temporarily cleared and restored afterwards). System language list: `de-DE, ja-DE`.

| Build | `CFBundleLocalizations` | Settings window |
|---|---|---|
| pre-fix | *entry does not exist* | **English** — the bug |
| this branch | `de en es fr it ja ko pt-BR ru zh-Hans zh-Hant` | **German** |

Owner confirmed both sides visually.

Two things this pins down that the unit tests cannot:

- **The failure is real in a packaged build and invisible outside one.** The identical pre-fix expression, run as a plain script on the same machine at the same moment, resolves `Locale.current` to `de_DE` and yields `de` correctly. It only answers English inside the `.app`, where the bundle's declared localizations constrain the resolution — which is exactly why `swift run` never showed it.
- **The array is derived, not typed.** Eleven locales, `en` among them because `en.json` is on disk as the translator manifest, so English joins by construction rather than by a special case.

Issue #663, opened 2026-08-02
🗺️ Roadmap: the road to 1.0


hajiboy95:
### Scope & guiding rule

Successor to #136 (waves 11-14) and #133 (the release &
distribution collector), both closed complete: the release chain
they existed to sequence — #89 `.app` bundle, #32 release workflow,
#116 shim sweep, #105 Homebrew cask, #117 install CTAs, #106
Cloudflare Pages — has shipped end to end. What is left is no
longer a *pipeline*, so it does not want a pipeline's shape.

**Guiding rule: 1.0 is a quality bar, not a feature list.** The
public surface froze with the release wave. Everything below is
either a defect against that surface, debt that lets a defect ship
unnoticed, or one of the two things a 1.0 is expected to have and
KiwiDesk does not yet. Nothing here adds a layout, a mode, or a
new configuration axis; a proposal that does belongs behind 1.0,
not in it.

Execution conventions carry over: cluster = worktree = branch =
PR; parallelize across file-disjoint lanes, serialize within a
lane; large architecture work gets a dedicated session.

**Legend:** 🟢 parallel-safe · 🔴 serialize · ⛔ blocked/deferred

**Milestone moved to Roadmap (2026-08-18).** This collector no
longer blocks a `v1.0.0` tag — the release workflow's gate refuses
only a milestone named for the version being cut. It stays open as
the collector. Whether it also *closes* depends on whether the
shapes parked in it are tracked elsewhere; closing it while they
are not would lose them.

---

## What is left — 7 open

The waves below are **historical**: every one of them is shipped
or moved. This section is the live list, grouped into the lanes
that will actually carry them (execution order in
`plan/road-to-1.0.md`). Ordered by what 1.0 cannot ship
without. Refreshed 2026-08-23 (evening): the 2026-08-22 sweep's
filings are all closed — #935/#936 merged as PR #947,
#938/#939 as PR #945, #940 as PR #946, #937 as PR #949 — along
with the same-day #944 (the track-weight heal, PR #948), and
#845 shipped and closed with PR #950 (the ⌃. row chord,
device-verified). The same evening's live QA then filed five:
#951/#952 (the close-return focus cluster — Ghostty's quick
terminal and the ⌃⌥K cheat sheet each steal focus on
dismissal), #953 (the tiled Settings window's mouse resize
moves no neighbors), and #955/#956 (the notification-red badge
default and the accordion affordance, each carrying a
[...]

Issue #531, opened 2026-07-25
test: layout tests read the real NSScreen, so host display size decides what they assert


hajiboy95:
Follow-up from #523. Not urgent — the suite is green on both the
runner and a dev Mac today. This is about *why* it is green.

## The coupling

Layout tests exercise `calculatedFrames`, which resolves its
bounds from the host's **real** display:

```swift
// TilingEngine+Layout.swift
let screen = NSScreen.main ?? NSScreen.screens.first
```

There is no seam to pass a rect in. 18 files under
`Tests/KiwiDeskCoreTests/` reference `NSScreen`; some derive their
expectations from it (`NSScreen.main?.visibleFrame.width ?? 1920`),
which is honest, and some assert a fixed arrangement, which is not.

So the same test can build a **different layout** on a dev Mac
(1728pt wide) than on the CI runner (~1066pt), with no code
difference between the two runs.

## What it already cost

This is not hypothetical — it bit the moment CI could run tests
again. With the default `min_window_size` of 300pt, a region below
`2 * min` can't hold a side-by-side pair, so `BspLayout` correctly
falls back to an `OverlapStack` pile (#383/#44). On the runner the
bottom-right BSP region is 533pt against a 600pt requirement, so
the pair piled and three `BspFocusReachabilityTests` assertions
failed; `StackArrangementResizeTests` failed the same way on the
short axis.

Worth knowing for the next time: `OverlapStack.offset` is 40pt and
vertical-only, so a pile shows up as **equal `minX` and midYs
exactly 40 apart**. And a CI-only geometry failure reproduces
locally not by chasing the screen but by **raising
`min_window_size` until the same threshold trips** — 500pt on a
1728pt screen reproduced all four failures exactly.

#523 fixed those two fixtures by pinning a 100pt minimum. That
bought headroom, not independence — every other geometry suite
still inherits whatever display the host happens to have.

## Options

**A — inject the bounds.** Give the layout path a testable seam
taking an explicit rect, so a fixture states the display it means.
Matches AGENTS.md §2.6 ("pure state and layout code must stay
actor-free and unit-testable") and §5 ("resolve before layout").
Cost: a production seam plus a sweep of the affected suites.

**B — a shared fixture constant.** Every geometry suite pins a
known-safe `min_window_size` from one place, so the threshold is
named once rather than rediscovered. Cheaper, but still leaves the
arrangement a function of the host screen — it only widens the
band where that doesn't show.

[...]

PR #667, opened 2026-08-02
feat(bar): ship the bars on the top and bottom edges, plain and opaque

### LINKED ISSUE #660: 💡 Retune the shipped bar defaults: Space Bar on top, App Bar on bottom, plain and less transparent
### Use case

The shipped bar defaults fight the macOS surfaces they sit beside.
Out of the box the Space Bar takes the **left** edge and the App
Bar the **top**, both in the `boxed` design language, and both
with a 40 %-opaque fill (`#14201C66`) that on a light or busy
wallpaper leaves them hard to read.

A new user should get a good arrangement with almost no
configuration (AGENTS.md §2.7 — approachable by default), and the
first-run impression should read as part of the system rather than
as two floating widgets pasted over it.

### Proposed solution

Change four shipped defaults. `SpaceBarStyle` and `AppBarStyle` are
the single source for all of them — `TilingSettings` only
instantiates the two structs, and the standard profiles carry no
bar styling of their own — so this is a change to the struct
defaults plus the docs and fixtures that quote them.

| Setting | Today | Proposed |
|---|---|---|
| `space_bar.edge` | `left` | `top` |
| `app_bar.edge` | `top` | `bottom` |
| `space_bar.background_style` | `boxed` | `plain` |
| `app_bar.background_style` | `boxed` | `plain` |
| `space_bar.fill_color` | `#14201C66` (40 %) | `#14201CCC` (80 %) |
| `app_bar.fill_color` | `#14201C66` (40 %) | `#14201CCC` (80 %) |

Top and bottom mirror the menu bar and the Dock, which is where
the eye already expects a persistent strip. `plain` reads as a
system surface rather than a floating plate — the same reason the
menu bar has no rounded box — and 80 % keeps a translucency
character while staying legible over any wallpaper.

### Alternatives considered

- **Leave transparency alone and rely on the palette.** Rejected:
  the problem is wallpaper-dependent, so no palette choice fixes it
  for everyone.
- **Go to 90 % (`#14201CE6`).** Near-solid; maximum legibility but
  it drops the translucency character the bars are designed around.
- **Make the bars opaque only in `plain`.** Two coupled defaults to
  explain, for a difference the user can set with one control.

### Extra context

Only the shipped defaults move — every one of these stays reachable
from Lua and from Settings, so anyone who prefers the current look
can set it back in six values. Pre-release, so no migration path is
needed: re-saving the config *is* the migration.

hajiboy95:
Closes #660.

| Setting | Was | Now |
|---|---|---|
| `space_bar.edge` | `left` | `top` |
| `app_bar.edge` | `top` | `bottom` |
| `space_bar.background_style` | `boxed` | `plain` |
| `app_bar.background_style` | `boxed` | `plain` |
| `*.fill_color` | `#14201C66` (40%) | `#14201CCC` (80%) |

Each half is the same argument. Top and bottom are where the menu bar and the Dock have already taught the eye to expect something permanent, so the bars read as part of the system rather than as two panels stuck over it; the left-edge Space Bar matched nothing. `plain` — one shared plate — is what the menu bar itself does, while a box per item reads as a widget floating on the desktop. And a translucent default is a bet on the user's wallpaper: 40% was legible on the dark ones it was chosen against and a guess everywhere else, whereas opacity is the one axis where the safe default costs a confident user exactly one setting.

Nothing is narrowed — all six stay reachable from Lua and Settings, and the struct defaults are their single source (`TilingSettings` only instantiates them; no standard profile carries bar styling).

## Fixtures, and one that was half-pinned

Three parity suites pick, per field, a value that must **differ** from the default — and `edge` and `background_style` had been picked as what are now the defaults. `backgroundStyle` has only two cases, so the global and override fixtures have to take one each.

`StackFocusResizeTests` is the interesting one: it pinned the display (#531) but not the bar, and the Space Bar carves its strip out of that frame before any layout sees it. That cost nothing while the bar sat on the *left* and the arithmetic was vertical — moving it to the top shifted the cliff cap by exactly 32/300. The suite now disables the Space Bar rather than re-pinning it to an edge.

## Two things I did not do, deliberately

- **The other ten `visibleBounds`-pinning suites were not given the same pin.** The full suite is green, so no defect is proven; applying it as churn to ten files would be a change nobody can verify. The obligation is captured in `.claude/rules/tests.md` instead (see #663's companion PR).
- **`AppBarStyle.swift` is now 321 lines**, over §2.1's 250 target and under the 350 ceiling, so `lint.sh` warns and the gate stays green. Splitting the colour block out is a real cleanup but not this change's job.

## Not eye-confirmed

`#14201CCC` is my number, not a measured one. You waived the on-device check to unblock the PR — worth a look before merge, since it is the one value here that only your eyes can settle.

## Verification

`swift build` · full `swift test` (2317 passed) · `scripts/lint.sh` — all green.


hajiboy95:
Device QA done (2026-08-02, macOS 26.6, M1 Max, signed build from this branch).

Geometry measured via `CGWindowListCopyWindowInfo` on a 1728x1117 main display:

```
Space Bar   y=32    1728x32   full-width, directly under the menu bar
App Bar     y=1077  1712x32   bottom, inset 8pt (the outer gap)
```

Owner confirmed on-device: placement good, and the 80% fill reads better than the old 40%. `plain` accepted in place.

Config was verified clean first — no `app_bar`/`space_bar` key anywhere in `gui.json`, `init.lua` or the `Starter` profile — so the struct defaults are genuinely what rendered, not a saved config.

Issue #636, opened 2026-07-29
borders: focus ring missing when open-or-focus switches spaces


hajiboy95:
## Repro (owner report, 2026-07-30)

1. Space 2 holds two windows: window A (the app a shortcut targets) and window B (another app).
2. Focus window B on space 2, then switch to space 1.
3. On space 1, press the app shortcut in **Open or Focus** (`pull_or_spawn`) mode.

**Result:** the space switches and window A receives focus, but no focus ring is drawn.
**Expected:** window A wears the focused ring after the switch, same as any other focus change.

The precondition in step 2 matters: when window A was the space's last-focused window, the ring appears.

## Analysis pointers

- `pull_or_spawn` is a fire-and-forget `NSRunningApplication.activate()` (`KiwiCore+Commands.swift`, `launch`); everything after it is reconstructed from racing async events (app-activated notification, AX focus report, space activation, post-switch re-tracking).
- `KiwiCore.borderSpecs` returns an **empty spec list** when the resolved focused window has no slot in the active space — zero rings anywhere, which matches the symptom exactly (no ring, rather than a misplaced one).
- Candidate staleness sources: the `.windowCreated` fold makes every (re-)tracked window its space's `focused` (`StateCoordinator.apply`), and the post-native-switch settle re-focuses `activeSpace.focused`.

Needs a state-level repro test to pin which ordering leaves the anchor stale; the fix is bookkeeping/ordering, not an OS limitation.

Issue #300, opened 2026-07-16
fix(borders): suppress focus border on launcher/Spotlight overlays


hajiboy95:
## Bug

When a launcher overlay (Apple Spotlight, Raycast, Alfred) grabs focus, KiwiDesk draws a focus ring around it.

## Why it happens

`borderSpecs(...)` (`KiwiCore+Borders.swift:96-104`) draws the focused ring **regardless of floating** — the `!floating` filter (`:108-109`) only suppresses *unfocused* rings. So a launcher that auto-floats (accessory activation policy / panel subrole / layer ≠ 0) but momentarily takes focus still gets a ring.

These overlays are **not** the phantom-window case that `isUnbackedAuxiliary` (`FloatDetection.swift:42-54`, #177) already handles — that catches AX proxies with **no** backing WindowServer window (`layer == nil`), e.g. the language-input popover. Spotlight/Raycast are *real, layered, focus-taking* windows, so they slip past that net.

## Recommended fix — a general heuristic, not a hard-coded ignore list

Suppress the focus ring when the focused window is a floating **transient overlay** (accessory activation policy, or non-standard panel subrole). No bundle-ID list to maintain; covers any launcher automatically.

Contrast with the #177 `ignoredSystemApps` list — that was a *narrow* belt for two text-input agents that needed app-wide ignore. Growing it per-launcher would be the wrong pattern here: these overlays float correctly and only the border is wrong, so the fix belongs at border-draw time, not in the management gate.

**Only reach for built-in ignore** (never-in-state) if we decide these overlays must not be tracked at all — likely unnecessary for a momentary overlay that floats correctly.

## Scope

- **v1:** the border-suppression heuristic above (Spotlight is the confirmed repro; Raycast/Alfred expected same).
- **v2:** a general user-facing ignore surface — ignore rules are Lua-only today (`GuiConfig.swift:43`, "no GUI control #176") — for anyone who wants a launcher fully unmanaged.

Relates #176, #177, #278 (borders).

PR #679, opened 2026-08-02
fix(state): stop a popup taking its space's focus when it spawns

### LINKED ISSUE #671: 🐛 Every fresh window steals its space's focus — popup windows warp the mouse on dismiss

0.9.4 (`640040ea`)

### macOS version

macOS 26.6

### Mac model type

### Steps to reproduce

1. Tile a space with Telegram plus other windows.
2. Right-click a message in Telegram — the context menu is a new
   AX window.
3. KiwiDesk grants it the space's focus on creation.
4. Dismiss the menu: the close-side handoff issues a real AX
   raise plus a mouse warp onto the fallback window.

Likely reproducible with any app whose popups surface as AX
windows, not only Telegram.

Expected: a transient/floating popup never takes the space's
focus at creation, and its destruction triggers no focus handoff,
no raise, no mouse warp.

Actual: `applyWindowCreated` grants focus to every fresh window —
`State/StateCoordinator+WindowCreated.swift:78-82` takes the
`!hadRememberedSpace` branch with **no classification signal
consulted**: `window.isFloating` is in scope at `:31` and
`isTransientOverlay` is on the model, neither is read. PR #640
deliberately narrowed only the *returning*-window case; fresh
spawns kept the unconditional grant. Because the popup became
`space.focused`, its destroy sets `focusLost` and
`App/KiwiCore+Events.swift:212-217` runs `focusWindow(next,
warp: true)` — `AXHelper.raise` sets `kAXMain`, performs
`kAXRaiseAction`, force-activates the app, and warps the pointer
(`AX/AXHelper.swift:287-302`). Collateral: the create also
cancels a pending legitimate focus-follow
(`App/KiwiCore+Events.swift:50`), and in focus-driven modes the
space pans toward the popup.

If a popup reports layer 0 + standard subrole it is additionally
fully tiled (`AX/FloatDetection.swift:10-29`) — a separate,
rarer failure of the same blind spot.

### Relevant output and logs

Not captured — static analysis with file:line evidence above.

- [x] None that I know of

### Extra context

Fix shape: deny the create-time focus grant to windows classified
transient/floating (`isTransientOverlay`, or at minimum
structural floats), and make the close-side handoff skip windows
that never legitimately held focus. Care needed around the #640
regression tests (`windowCreated` fold) — the returning-window
behavior must stay.

The upstream gate must NOT be widened instead: regular apps'
panels "must only float" per
`Events/EventLoop+WindowPolicy.swift:84-90` — this is a
focus-grant fix, not an adoption filter change.

Roadmap: #663 Wave 1, P1.

hajiboy95:
## Description

`applyWindowCreated` granted the space's focus to **every** fresh
window, with no classification signal consulted. A popup that
surfaces as an AX window — a Telegram right-click menu, any app's
raised-layer panel — therefore became `space.focused` the moment
it appeared, and its dismissal read as the focused window
closing. The fallback handoff then fired for real: a
`kAXRaiseAction` that re-activates an app, and under
mouse-follows-focus a pointer warp off what had just been
clicked. In a focus-driven layout the grant also panned the space
toward the popup.

The fix is one condition in the create fold: a window classified
as a **transient overlay** (`ManagedWindow.isTransientOverlay` —
the #300 structural-float classification) is granted nothing.

**Only the grant — never the slot.** A window in this class that
macOS genuinely focuses still lands in `space.focused` through
the focus report a moment later. That matters because the same
flag also covers layer-0 dialogs and accessory apps' settings
windows, which #300 explicitly ruled behave correctly and want
their focus, with only their ring wrong. An earlier pass of this
branch barred the whole class from the slot; review caught that
it would have put every focused command on the window behind the
one being typed in, and it was withdrawn.

The signal is deliberately the structural overlay flag, not
floating-ness: a window the user floated through a `float_rules`
entry is ordinary and still takes focus when it spawns. The flag
is read back from state rather than the incoming snapshot,
because a remembered-tiled restore in the same fold clears it.

## Related Issue(s)

Closes #671

## Checklist

  (see AGENTS.md §3)
  required)
  passes

## How I verified

Static analysis plus the state-level suite; **the on-device
eye-confirm is still owed** — the repro is a Telegram (or any
app's) right-click menu in a tiled space: dismissing it must not
raise another window or move the pointer.

`swift build`, `swift test` (2338 tests / 421 suites),
`scripts/lint.sh` and the site build are green. The release
build was not run locally — no concurrency or `Sendable`
surface is touched — so CI's `Release Build` job is the one to
read before merging.

`Tests/KiwiDeskCoreTests/TransientOverlayFocusTests.swift` (7
tests) pins both arms: the overlay spawn taking nothing and its
dismissal reporting no focus loss, against a `float_rules`
window and an ordinary window that still do take focus. Two more
pin the boundaries this fix deliberately does not cross — an OS
focus report still moves the slot, and #636's returning-window
seed still fires. `guard-prover` red-proofed the suite (an
earlier revision; the current one is a strict narrowing of it).

## Notes for Reviewer

Three commits, and the middle one is a design that was tried and
withdrawn — worth reading the third commit message rather than
the diff of the second. The net production change against `main`
is a single condition.

Two follow-ups were raised in review and deliberately left out
of scope: the pointer warp is keyed on the window id rather than
the slot (so `mouse.follows_focus` can still warp onto an
overlay macOS focuses), and `emitFocusChange` names that window
while `space.focused` names the one underneath. Both are
pre-existing and unchanged by this PR.


hajiboy95:
CI is green, **including the `Release Build` job** (1m48s) — the checklist box that was left unticked pending it can be considered satisfied. `Build, Lint & Test`, `Build site` and Cloudflare Pages all pass.

Still genuinely outstanding: the **on-device eye-confirm**. Repro is a right-click context menu (Telegram, or any app whose popups surface as AX windows) in a tiled space — dismissing it must not raise another window or move the pointer.

Issue #418, opened 2026-07-21
💡 Floating windows always on top of tiled windows


hajiboy95:
## Use case

When mixing floating and tiled windows on the same space, a floating
window should act as an "always on top" layer — visible above every
tiled window regardless of which tiled window is focused. Today,
clicking (or cmd+tabbing to) a tiled window naturally raises it via
macOS's own stacking, burying the floating window behind it. The user
must manually re-focus the float to bring it back.

**Expected behavior:**
- Floating windows always render above all tiled windows, even when a
  tiled window is focused.
- When two floating windows overlap, the last-focused float sits on
  top (natural most-recently-focused stacking within the float layer).
- The `isFloating` flag is the gate — not `isSticky`. A window that
  is both floating AND sticky gets the level promotion (it floats on
  every space). A window that is sticky but NOT floating tiles
  normally after #414 v2.

**No toggle needed.** Floating inherently means "above the tiled
plane" — that's the semantic distinction from tiled. A user who wants
a window excluded from tiling but NOT pinned above others uses
`ignore_rules` (KiwiDesk doesn't touch ignored windows at all,
including z-order). The three tiers are:

| Tier | Z-order | Managed by KiwiDesk |
|---|---|---|
| **Tiled** (incl. tiled-sticky) | Normal window level | Yes — layout, frame, z-order |
| **Floating** (incl. floating+sticky) | Above tiled | Yes — free-positioned, level-promoted |
| **Ignored** | Untouched | No — invisible to KiwiDesk |

## Proposed solution

Two-tier approach matching the project's private-API guardrail
(AGENTS.md §5: every private fast path must have a public-API
fallback).

### Fast path: `SLSSetWindowLevel` (private SkyLight)

Resolve `SLSSetWindowLevel(connection, windowID, level)` via `dlsym`
in `SkyLight.swift`. When a window becomes floating
(`setFloating(_, true)` / float-rule match / `make_floating`
command), set its window level to `kCGFloatingWindowLevel` (3) — the
same level macOS uses for its own floating panels. When it returns to
tiled (`setFloating(_, false)` / `make_tiled`), restore it to
`kCGNormalWindowLevel` (0).

This is instant, survives focus changes (the window server enforces
the level), and requires no per-focus-event re-raise.

**Within the floating level**, z-order is natural macOS stacking:
the last-focused float is on top automatically — no extra work
needed. A `raiseQuietly` at make-floating time ensures the newly
[...]

Issue #153, opened 2026-07-10
pendingZOrderRestore armed before a retile can fire mid-retile, ahead of the new animations


hajiboy95:
**Finding** (code review of #143, noted out-of-scope there;
same mechanism as the #143 fix-batch reorder). `TilingEngine.
retile` cancels in-tolerance windows one by one
(`animation.cancel` → `notifyIfIdle`), so
`onAllAnimationsEnded` can fire **synchronously mid-retile**
the moment the count transiently hits zero — before the same
retile's `animate` calls start the new animations. Frame
iteration order is a dictionary, so this is timing/order
dependent.

`focusWindow` now arms `pendingFocusRaise` strictly AFTER the
retile for exactly this reason (#143 review fix). But
`pendingZOrderRestore` schedulers that run before a retile —
e.g. `KiwiCore+NativeSpaces.swift` scheduling around a space
apply — can still have the restore fire in that mid-retile
window, re-raising piles from pre-retile state and then being
consumed, so the intended post-settle restore never runs.

**Fix sketch.** Audit `scheduleZOrderRestore` call sites for
schedule-before-retile ordering and move them after the retile
they pair with (the flag is one slot, so this is per-call-site
reordering, not a mechanism change). Alternatively make
`retile` suppress settle notifications for its own duration
(begin/end batching in `AnimationEngine`) — heavier, only
worth it if call-site ordering proves fragile.

Found by the #143 review; refs #143.

hajiboy95:
Audit result from the #143 round-2 architect review (2026-07-10),
recorded so the implementer skips a re-audit:

- Already arm **after** their paired retile (no action):
  `KiwiCore+Drag.swift:215`, `KiwiCore+NavigateCommand.swift:96`,
  `KiwiCore+NativeSpaces.swift:94` (the site the issue body
  names as an example is in fact fine).
- The one confirmed arm-**before**-retile site is
  `promoteDemote` (`KiwiCore+StackCommands.swift:141`), and it
  can NOT be fixed by reordering within the call site: its
  paired retile is the *dispatcher's* trailing
  `retile(force: true)` in `KiwiCore+LayoutCommands.swift:34-41`.
  So the fix is a design choice, not a two-line reorder —
  either a "schedule after my retile" signal across the
  command/dispatcher seam, or `AnimationEngine` begin/end
  settle batching (the heavier option the fix sketch mentions).

hajiboy95:
[...]

Issue #150, opened 2026-07-10
Scrolling swap never schedules a z-order restore for the edge piles


hajiboy95:
**Finding** (code review of #147, nit — pre-existing, not a
#147 regression). `restoreScrollingZOrder`
(`KiwiCore+ZOrder.swift:68`) exists to re-stack scrolling's
edge piles (nearest-to-focus on top), but its only schedulers
today are stack-boundary crossings, native-space switches,
mode switches, and profile loads. A `swap` in a scrolling
space — geometric path and the #147 array step alike — never
calls `scheduleZOrderRestore()`, so a swap that pushes the
ex-neighbor into an edge pile can leave the pile with stale
stacking (a farther column sitting on top of a nearer one).

**Fix sketch.** Schedule the restore from the scrolling swap
paths (both `scrollingStep` and the geometric swap when the
active mode is scrolling), mirroring how stack swaps schedule
on zone crossings. The existing settle-deferral (`pending`
flag + `onAllAnimationsEnded`) already handles the timing.
Consider only scheduling when either window's slot is outside
the viewport (an all-visible swap scrambles nothing).

Found by the #147 review; refs #147, #142.

hajiboy95:
Shipped in PR #159 (merged as d5eda7f); the PR body listed this fix but had no closing keyword, so the issue stayed open. Both scrolling swap paths (the #147 array step in `scrollingStep` and the geometric swap in `navigate`) now call `scheduleScrollingZOrderRestoreIfOverflowing()` after their retile, gated on `ScrollingLayout.rowOverflows` (a row that fits the viewport has no piles to scramble). Regression tests in `ZOrderScheduleTests.swift`. Verified on main today: 1074 tests pass.

PR #680, opened 2026-08-02
fix(zorder): restack a scrolling row after an App Bar jump or bar drop

### LINKED ISSUE #674: 🐛 Scrolling z-order restore is never armed on App Bar click jumps or bar-item drags

0.9.4 (`640040ea`)

### macOS version

macOS 26.6

### Mac model type

### Steps to reproduce

1. Scrolling layout with an overflowing row (edge piles on both
   sides).
2. Click an App Bar item several slots away from the focused
   window (a bar click jumps an arbitrary gap — the item carries
   an absolute `WindowID`).
3. The target raises, but the windows between old and new focus
   keep the stale stacking implied by the previous focus — the
   edge piles read wrong.
4. Same after dragging a bar item across several slots: the
   reorder retiles but never restacks.

Expected: after a multi-slot focus jump or a bar-item reorder,
the row is restacked canonically (both sides,
farthest-from-focus first, focused last) — exactly what
`restoreScrollingZOrder` already produces.

Actual: the restore machinery is never armed on either path.
`focusWindow` (`Commands/KiwiCore+FocusRaise.swift:73-166`)
raises only the target and arms `scheduleZOrderRestore()` solely
for `.monocle` (`:147-149`). The scrolling restore
(`scheduleScrollingZOrderRestoreIfOverflowing`,
`Commands/KiwiCore+ZOrder.swift:35-44`) is armed only by swaps
(`KiwiCore+NavigateCommand.swift:186`, `:273`), window drags, and
native space switches. `moveBarItem` ends with a bare `retile()`
(`App/KiwiCore+AppBar.swift:213`) — the same array mutation as
`scrollingStep`'s swap, which does arm it. A ±1 step masks the
gap because the lone target raise suffices; a gap jump does not.

### Relevant output and logs

Not captured — static analysis with file:line evidence above.

- [x] None that I know of

### Extra context

Fix is two arm calls:

1. In `focusWindow`, arm
   `scheduleScrollingZOrderRestoreIfOverflowing()` after the
   retile (the #153 ordering rule — "call after the swap's
   retile", `KiwiCore+ZOrder.swift:29-34`).
2. Same at the end of `moveBarItem`.

No per-anchor-side logic needed: the restore only changes
z-order, never positions, so a side already stacked canonically
shows zero visual change; the existing `rowOverflows` gate
already skips all-visible rows. Must order against the 50 ms
coalesced float re-raise
(`Commands/KiwiCore+ZOrderFloats.swift:89-105`).

Tests: extend `ZOrderScheduleTests` (`:73`, `:91` cover the swap
path; no equivalent for a focus jump or a bar drop) — new guards
get a `guard-prover` run.

Roadmap: #663 Wave 1, P2. Smallest chunk of the batch.

hajiboy95:
## Description

The scrolling edge-pile z-order restore was armed only by swaps,
window drags and native space switches. A ±1 focus step masked
the gap — the lone target raise is the whole visible change — but
an App Bar item carries an **absolute `WindowID`**, so a click can
cross several slots at once and every window between the old and
the new focus keeps the stacking the old focus implied. The edge
piles then read wrong. Dragging a bar item had the same hole:
`moveBarItem` ends in a bare `retile()`, the same array mutation
as `scrollingStep`'s swap, which *does* arm the restore.

Two arm calls, both after their retile (the #153 ordering rule).

**Both are narrow, and the narrowness is load-bearing.** The
first pass of this branch armed on every scrolling focus change,
arguing the cost was nil because a canonically-stacked side
re-raises to the order it is already in. Review showed the cost
is not pixels: `restoreScrollingZOrder` raises every tiled
pile-mate through the blocking ordered queue, which leaves the
tiled plane above the float layer until the next genuine focus
event (#418), and drags ring and mark a restack behind while the
queue drains. So the focus arm now fires only when the focus
crosses **more than one** tiled slot, and not at all when either
end is missing from the row (a first focus, or one arriving from
a focused float) — without both indices there is no evidence a
jump happened, and guessing costs more than waiting.

The bar-drop arm fires only for the **active** space: the whole
restore path reads `activeSpace`, while a bar belongs to the
space showing on its own display, so a drop on a second display's
bar would have restacked a space the user is not in and left the
dropped row stale anyway. The secondary-display half of #674 is
left as a separate decision, recorded at the call site.

## Related Issue(s)

Closes #674

## Checklist

  (see AGENTS.md §3)
  required)
  passes

## How I verified

Static analysis plus the new suite; **the on-device eye-confirm
is still owed** — a scrolling space with an overflowing row (edge
piles both sides), then an App Bar click several slots away, and
a bar-item drag across several slots. The piles should read as
the row receding from the new focus in both directions.

`swift build`, `swift test` (2338 tests / 421 suites) and
`scripts/lint.sh` are green. No docs page changed, so no site
build was needed. The release build was not run locally — no
concurrency or `Sendable` surface is touched — so CI's `Release
Build` job is the one to read before merging.

`Tests/KiwiDeskCoreTests/ZOrderFocusJumpTests.swift` (7 tests)
pins the two arms and, against them, the three cases that must
not arm: a single-slot step, an all-visible row, and a drop on an
inactive space's bar. `guard-prover` red-proofed the first
revision (5/5, no inert assertions); the inactive-space test was
rewritten after review found it proved the mode gate rather than
the guard it named, and it now gives the active space its own
overflowing scrolling row.

## Notes for Reviewer

Two commits: the first is the plain two-arm fix from the issue,
the second is the narrowing that review earned. The second
commit's message carries the reasoning.

`.claude/rules/state-and-layout.md` gains the arming obligation
with a §5 row — three mutations have now shipped without an arm
(#150, #153, #674), and nothing scans for a fourth.


hajiboy95:
CI is green, **including the `Release Build` job** (2m30s) — the checklist box left unticked pending it can be considered satisfied. `Build, Lint & Test` and the change-detection job both pass.

Still genuinely outstanding: the **on-device eye-confirm**. Repro is a scrolling space with an overflowing row (edge piles both sides), then an App Bar click several slots from the focused window, and a bar-item drag across several slots — the piles should read as the row receding from the new focus in both directions.

Issue #684, opened 2026-08-02
🐛 Pile raises land out of order: AXRaise returns before slow apps perform it


hajiboy95:
### KiwiDesk version or commit

0.9.4 (`71108293`)

### macOS version

macOS 26.6

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Scrolling space, 8 windows, row overflowing (edge piles both
   sides). Include an Electron app (Obsidian) and Telegram.
2. Trigger a z-order restore — an App Bar click several slots
   away, a directional `swap`, or a bar-item drag.
3. Compare the resulting front-to-back order against
   `KiwiCore.scrollingRaiseOrder`'s.

Measured with a `CGWindowListCopyWindowInfo` poller at 50 ms,
logging layer-0 owners on every change.

### Expected vs. actual behavior

Expected: the restore's raise sequence is issued in order and the
final stacking matches it. `performZOrderSequence` states the
contract outright — "Raises run sequentially on one queue: each
call returns only after the target app processed it, which keeps
the order across apps."

Actual: **the order KiwiDesk computes is correct every time; some
apps perform the raise after the call returns**, so their window
lands out of sequence. `AXUIElementPerformAction` does not
guarantee the action completed on return, and Electron/WebKit
apps answer AX lazily (the 100–300 ms this repo already documents
for `AXEnhancedUserInterface` and the #465 re-report distrust).

Array order was `Claude(0) Notizen(1) Zen(2) Telegram(3)
Ghostty(4) ChatGPT(5) Fotos(6) Obsidian(7)`.

Jumping **to** Obsidian (focus index 7 — Obsidian is excluded
from the sequence and re-asserted last):

```
expected  Obsidian > Fotos > ChatGPT > Ghostty > Telegram > Zen > Notizen > Claude
actual    Obsidian > Fotos > Telegram > ChatGPT > Ghostty > Zen > Notizen > Claude
actual    Obsidian > Fotos > Telegram > Ghostty > ChatGPT > Zen > Notizen > Claude
```

Six of eight exact; only Telegram is too far forward — raised
4th, but the two raised after it failed to land above it.

Jumping **to** Claude (focus index 0 — Obsidian is raised FIRST,
so every other raise must land on top of it):

```
expected  Claude > Notizen > Zen > Telegram > Ghostty > ChatGPT > Fotos > Obsidian
[...]

Issue #635, opened 2026-07-29
site: branded trilingual 404 page (4🥝4) + shared AA green-text literal


hajiboy95:
## Current state

The docs/marketing site (`site/`, Astro + Starlight) ships Starlight's stock 404: generic "Page not found. Check the URL or try using the search bar." — English only, no brand identity — while the rest of the site is fully trilingual (en/de/ja) and the i18n JSON already exists under `site/src/i18n/`.

## Change

1. **Override the 404** with a real `site/src/pages/404.astro` using the family signature established on kiwicanopy.com: brand mark, a big **4🥝4** (kiwi emoji as the zero), playful one-line message, and a back-home link. English leads; German and Japanese follow as compact lines, since static hosting serves a single root `404.html` for all paths. Reference implementation: `KiwiCanopy-web/src/pages/404.astro` (commits `df30eec` + `8057e65`), including its theme-aware link treatment (light mode: ink text with accent underline for the primary link, darkened-kiwi `#506c37` for secondary links; dark mode: plain accent green).
2. **Unify the AA green-text literal** while in `site/src/styles/theme.css`: it currently uses an independently derived `#546b32` (≈5.6:1) for light-mode accent text. Adopt the family-wide `#506c37` (the 50/50 midpoint of `#8db354` and `#12251a`, 5.6:1 on `#f7f9f5`, documented in KiwiCanopy-web) so the three repos share one value for the same decision. Visually near-invisible.

## Context

Cross-repo brand audit (2026-07-29): greens are already byte-identical across KiwiCanopy-web, KiwiCV and KiwiDesk; the 404 experience is the main inconsistency, and this site is the weakest of the three while having every ingredient (i18n strings, correct theme tokens) already in place. KiwiCV keeps its compass NotFound deliberately — the 4🥝4 motif is the marketing-site signature, not an app-side requirement.

Strings needed (add to `site/src/i18n/*.json`, mirroring KiwiCanopy-web's `notfound_*` keys): title, message, back-home label × en/de/ja.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #696, opened 2026-08-02
fix(events): honor a click on a raise-stamped window (#687)

### LINKED ISSUE #687: 🐛 A click during a z-order restore focuses the window in macOS but not in KiwiDesk

0.9.4 (`c73af12c`, the #684 branch; the revert path itself is
untouched by it and the behavior predates it)

### macOS version

macOS 26.6

### Mac model type

### Steps to reproduce

1. Scrolling space, row overflowing so both edges pile.
2. Rearrange a window past the focused one — drag its App Bar
   item from left-of-focus to right-of-focus, or the reverse.
   Any action that arms a z-order restore will do; the drag is
   just the reliable one.
3. Within about a second of the restack settling, **click** a
   different window in the row.

Expected: the click focuses that window — ring moves, the row
pans to it.

Actual, intermittently: **macOS focuses the window but KiwiDesk
does not.** Keystrokes go to the clicked window, while the focus
ring stays on the previous window and no pan happens. The two
disagree until the next focus event. Clicking elsewhere and back
fixes it, which is the tell: the first click is consumed.

Keyboard focus commands are never affected — they do not produce
the AX focus report this path inspects.

### Relevant output and logs

The mechanism, not a log. A z-order restore stamps every window
it is about to raise in `zOrderRaiseEchoes` (#418/#425), because
`kAXRaiseAction` couples with app activation and emits a focus
echo carrying no self-raise provenance (#152). The revert in
`KiwiCore+FocusEvents.swift:75` then treats the first focus report
from a stamped window as that echo:

```swift
if let stamp = zOrderRaiseEchoes[id],
    Date().timeIntervalSince(stamp) < Self.zOrderRaiseEchoWindow,
    let intended = effects.focusBefore, intended != id,
    !outstandingSelfRaises.contains(id)
{
    zOrderRaiseEchoes[id] = nil
    if let space = state.workspaces.space(of: intended) {
        state.workspaces.focus(intended, in: space)
    }
```

A genuine click is shaped exactly like that echo, so it is
reverted and the stamp consumed — hence the second click working.

**Two distinct defects live here.**

1. *A real click is indistinguishable from a raise echo.* This is
   the known trade documented in `docs/accepted-limitations.md`
   (the scrolling backward-focus row).
2. *The revert only restores KiwiDesk's state, never the OS
   focus.* It calls `state.workspaces.focus(intended,…)` and
   stops, so after a wrongly-reverted click the clicked window
   keeps macOS key focus while KiwiDesk believes focus is
   elsewhere — the ring, the pan and every focused command act on
   the other window. That divergence is what makes this read as
   broken rather than as a missed click, and it is a bug on its
   own terms even where the revert is correct.

Fix sketch, using machinery that already exists: the #465 branch
immediately below this one already escapes on
`recentClickInside(id)` (`KiwiCore+FocusEvents.swift:265`) —
"unless a fresh click landed inside the reported window". The same
escape belongs on the raise-echo revert: a click *inside the
reported window* is provenance no raise echo can forge, since a
restore's echoes come from windows the user did not click. Worth
checking against the #431 traveler case and the float raise, both
of which lean on this revert, and worth deciding whether the
revert should re-assert OS focus rather than only state.

- [x] None that I know of

### Extra context

**Pre-existing, not introduced by #684** — that branch does not
touch `KiwiCore+FocusEvents.swift`. It does *widen the window*:
a raise sequence there verifies each landing against the
WindowServer, so it now spans ~50-400 ms rather than ~10 ms, and
a click is exposed for as long as the sequence runs. #684 also
removed one contributor in the same area — it releases the stamps
of windows a diffing restore decides not to raise, which
otherwise sat unconsumed for the full second.

Found during device QA of #684 (owner, 2026-08-02); split out of
that branch deliberately, since the fix is in focus-event routing
rather than in z-order.

hajiboy95:
## Description

A z-order restore stamps every window it raises so the raise's
focus echo can be reverted (#418/#425). A genuine click on a
stamped window is shaped exactly like that echo, so the first
click after a restack was consumed: macOS focused the clicked
window while KiwiDesk reverted its state — keystrokes went to the
clicked window while ring and pan stayed behind (the asymmetry
that makes #687 read as broken). #684 widened the exposure from
~10 ms to the drain's 50–400 ms.

Fix: a click that **reached** the reported window is provenance
no echo can forge, and it escapes the revert. Two load-bearing
choices:

- **Reached, not contained.** Edge-pile frames overlap, so a slow
  pile-mate's late echo can contain the click point too; honoring
  it would pan the row onto a window the user never clicked.
- **Resolved at press time.** The frontmost managed window at the
  point at press instant is exactly the window the press lands
  in. Echo-time resolution was built first and withdrawn in
  review: mid-drain, a quiet raise of a *same-app* sibling
  becomes that app's new key window, so a stamped sibling could
  climb above the clicked window and forge the escape — and a
  retile can move frames under a click up to 1 s old. The press
  stamp (`KiwiCore+Lifecycle`) resolves `reached` through one
  WindowServer stacking read (~0.4 ms, the #684 measurement) via
  the new `stackingOrderProvider` seam (nil in unit tests, the
  `frontmostPIDProvider` pattern).

Issue #687's second defect — the revert restores state without
moving OS focus — is decided explicitly as **decline to
re-assert**: OS focus during a sequence is owned by the
sequence's closing re-assert (the `performZOrderSequence`
completion), an echo landing after it finds OS focus already
restored, and the one case where the divergence persisted was
the wrongly-reverted click this PR removes. Argued in
`docs/design-decisions.md` ("Raise-echo revert: state-only, and
a click is provenance").

Accepted narrowness (documented + pinned): resolution skips
untracked windows, so a click on a non-click-through ignored
window resolves to the managed window beneath — fails toward
honoring a focus report, never toward eating one. A deliberate
*clickless* focus of a stamped window (cmd-tab, app-driven)
inside the ~1 s echo window is still eaten once, as before.

`KiwiCore.swift` hit the 350-line ceiling; the computed
accessors `configURL` / `defaultSocketPath` moved to
`KiwiCore+Accessors.swift`.

## Related Issue(s)

Fixes #687

## Checklist

  (see AGENTS.md §3)
  required)
  passes
  before merging; it reports, it does not block)

## How I verified

- Full gate: `swift build`, `swift test` (2433 tests / 439
  suites green), `scripts/lint.sh` exit 0, site build green
  (docs touched).
- guard-prover, two rounds, 15 mutations total, 0 inert: the
  revert call site, every condition of `recentClickReached` and
  `clickReachedWindow`, the stamp-survives contract, and both
  production wiring lines (`ClickProvenanceWiringTests` reds on
  wiring deletion while every behavior suite stays green — the
  #684 "helper watched, wiring untested" hole, closed).
- Device QA on the signed build is next and will be reported on
  this PR before merge: repro is an App Bar drag to rearrange an
  overflowing scrolling row, then a click on a different window
  within ~1 s — ring and pan must follow the first click;
  keyboard focus was never affected, and that asymmetry is the
  diagnostic, so both get tested.

## Notes for Reviewer

- Review ran per AGENTS.md §3: code-reviewer + architect-reviewer
  + docs-steward round 1, fixes, sequential re-review — final
  round zero findings. The one echo-time → press-time rework came
  out of the architect's round-1 major.
- The wiring guard's textual limits are shipped in its docstring
  rather than denied (the #635 practice).


hajiboy95:
Device QA (owner, 2026-08-03) — passed after one round of fixes found by it:

- The original #687 repro (bar drag → click within 1 s) passed on the first build; QA then surfaced the **next eater in the same path**: the scrolling restore's closing re-assert raises the already-key window, which emits no echo, so its `outstandingSelfRaises` entry never expired and ate the next direct click on that window (state/OS focus split, indefinitely — bar clicks masked it). Traced on device with branch-level logging.
- Fixed in the third commit: the self-echo classification is age-bounded via `selfRaiseStamps` (the last unbounded ledger), a fresh entry stands down for click provenance, and a press inside a painted bar strip resolves no window (the trace showed `reached` resolving *through* the App Bar).
- Re-QA on device: 20 direct clicks honored across Antigravity and Notes, restore echoes still reverted, zero self-echo eats.
- guard-prover round 3: 6 mutations, all arms red including the call sites; the one uncovered arm it found (Space Bar absorption) got its own test, red-proved inline.
- Full gate green after the round: 2438 tests / 441 suites, lint 0, site build 0.

Issue #688, opened 2026-08-02
🐛 The quit-grid restack issues its raises unverified, and cannot self-heal


hajiboy95:
### KiwiDesk version or commit

0.9.4 (`d103f6f9`, the #684 branch)

### macOS version

macOS 26.6

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. `quit.set_layout("grid")` with enough windows that cells pile
   — several apps, at least one slow to answer AX (Electron,
   Telegram).
2. Quit KiwiDesk.
3. Compare the resulting stacking against the circle
   `QuitGridLayout.raiseOrder` computed.

Not yet measured on device — filed from code inspection during
#684, where the identical failure *was* measured on the live
restore paths.

### Expected vs. actual behavior

Expected, and what `docs/lua-reference.md` promises today: "the
final stacking is deterministic: within a pile every title bar
stays visible, later cells sit above earlier ones, and whichever
window had focus at quit makes no difference."

Actual, expected to be intermittently wrong for the same reason
#684 was: `KiwiCore+Teardown.swift:269` issues the whole circle as
a bare loop of `AXHelper.raiseQuietly`, and that call returns once
the app has *accepted* the raise, not once it has performed it
(measured for #684: returns in 0.4-3.8 ms, window moves 1-20 ms
later). A loop therefore issues every raise while none of them has
happened, and the apps land them in whatever order they reach
them.

**This one cannot self-heal.** Every other restore family is
allowed to be late because the next restore diffs against reality
and fixes it — that is the argument #684 rests on for not
retrying. Here the raises are the last thing that happens before
management stops, so a miss is the arrangement the user is left
looking at.

### Relevant output and logs

The violated obligation, added in #684 and already naming this
site (`.claude/rules/state-and-layout.md`):

> Several raises that must land in a given ORDER go through
> `raiseSequentially` / `performZOrderSequence` — never a loop of
> bare `AXHelper.raiseQuietly` calls.

The teardown loop predates that rule and was left out of #684's
scope deliberately: #684 converted the four restore families that
share `raiseSequentially`, and teardown is a different path with
its own 1 s wall-clock budget and its own 0.25 s AX messaging
timeout.

### Other window managers or AX tools running?

- [x] None that I know of

### Extra context

Two things to decide together, which is why this is one issue:

1. **Whether to route teardown through `ZOrderDrain`.** It would
   verify each landing against the WindowServer before the next
[...]

PR #698, opened 2026-08-03
fix(zorder): retire the monocle arm's counter, re-fire held warps (#689)

### LINKED ISSUE #689: 🐛 zOrderRestoresInFlight now holds ~40x longer, dropping monocle restacks and mouse warps

0.9.4 (`c330de42`, the #684 branch)

### macOS version

macOS 26.6

### Mac model type

### Steps to reproduce

Monocle half:

1. A monocle space, several windows, at least one app slow to
   answer AX.
2. Click an App Bar item, then click another one within ~120 ms
   (a normal double-target correction, or held focus-nav).
3. The second restore is dropped.

Mouse-warp half, needs `mouse.set_follows_focus(true)`:

1. Any layout with an overflowing pile or a floating window.
2. Trigger a restore, then change focus by a route that warps at
   report time (cmd-tab, a deferred focus-follow) while the
   sequence is still draining.
3. The pointer does not follow that focus change, and nothing
   re-issues the warp afterwards.

`zOrderRestoresInFlight` counts z-order sequences that have not
re-asserted focus yet. Two unrelated behaviors gate on it:

- `KiwiCore+FocusRaise.swift:147` — the monocle restore arms only
  when the counter is zero, so a restore's own closing re-assert
  cannot re-arm it and loop. The gate keys on mode alone, so this
  counter is currently its only loop guard.
- `KiwiCore+MouseWarp.swift:45` — `mouseWarpEligible` returns
  false while the counter is non-zero, so pile raises churning
  focus cannot drag the pointer around (#186).

Before #684 a sequence was fire-and-forget and held the counter
for roughly the length of its AX calls — a few milliseconds. #684
made the sequence verify each landing against the WindowServer,
so it now holds the counter until the sequence settles: bounded
at `ZOrderDrain.landingLimit` per window and `totalLimit` per
sequence, but ~40x the old window and, unlike before,
deterministic.

**What makes this worth filing rather than accepting quietly is
that two documented residues make the long hold routine on
exactly the paths these gates guard, not exceptional:**

1. The monocle restore's floor contains the frontmost app's key
   window on the App Bar path it exists for — and a quiet
   `AXRaise` cannot lift a window above the key window (measured
   for #684: 7 windows, none displaced, over 600 ms). So its
   landing can never verify there and the drain pays the full
   `landingLimit` before giving up. The raise is still issued
   exactly once, as before, so there is no functional regression
   in the restore itself — but the counter is then held ~120 ms
   after *every* monocle bar click, which is precisely when a
   user clicks again.
2. The mixed-layer float residue documented on `floatLayerTargets`
   spends `landingLimit` per float raise in that configuration,
   and float raises fire on every genuine focus onto a tiled
   window.

### Relevant output and logs

Not a crash; a dropped restack and a missing warp. The mechanism
is above.

- [x] None that I know of

### Extra context

**Deliberately not fixed in #684**, on review advice, because the
correct fix is not a narrowing of the counter but a replacement
of what the monocle arm tests. The scrolling arm one function
below already argues the better shape: it refuses the restore's
own re-assert *semantically* — the re-assert targets the focus
that is already current, so the jump distance is zero — rather
than by asking whether a restore happens to be in flight. Giving
monocle the same semantic test would remove its need for the
counter entirely, and that is a change with its own reasoning and
its own red-proof, not a line bolted onto a z-order ordering fix.

The mouse-warp half is a separate decision: either bound the warp
bracket independently of the drain's budget, or re-issue the warp
when the counter drops.

Note for whoever picks this up: the comment at
`KiwiCore+FocusRaise.swift:147` still describes the gate as
protecting against a re-entrant loop, which reads as rare. Since
#684 it also suppresses a deterministic ~120 ms window after
every monocle restore. Update that comment with whatever lands,
or the next reader inherits the old model.

Raised by `architect-reviewer` and `code-reviewer` independently
during #684's review; both agreed the fix belongs outside that
branch and the fact does not.

hajiboy95:
## Description

`zOrderRestoresInFlight` counts z-order sequences that have not
re-asserted focus yet. Before #684 a sequence held it for a few
milliseconds; the verified drain holds it 50–400 ms,
deterministically, and two behaviors gated on it broke on exactly
the paths they guard:

- **The monocle restore arm** keyed on the counter as its loop
  guard. A monocle restore's floor contains the frontmost app's
  key window — which a quiet raise can never clear — so the drain
  always pays its per-window limit (`ZOrderDrain.landingLimit`),
  and the counter then suppressed a **genuine** second restore
  after every monocle bar click, precisely when a double-target
  correction or held focus-nav clicks again. The arm now refuses
  **semantically**, the scrolling arm's shape (#674) and what
  both #684 reviewers prescribed: the restore's own closing
  re-assert targets the focus that is already current, so
  `previousFocused != id` refuses it by construction. Known
  residue (tiled-sticky traveler holding system focus arms one
  extra empty-plan restore, completion-gated, cannot loop) argued
  in the comment; architect-reviewer verified the termination
  argument against both completion paths.

- **The mouse warp** was dropped outright by `mouseWarpEligible`
  while the counter held, and nothing re-issued it. The hold
  moves into `warpMouseToFocused`: a warp landing mid-drain is
  recorded (`pendingMouseWarp`, single slot, latest-wins) and
  re-fired when the **last** overlapping drain ends —
  staleness-checked at fire time (the `runPendingFocusRaise`
  pattern), so maintenance churn still never drags the pointer
  and a stale target is dropped. The warp's machine tail moves
  behind a `pointerWarp` seam (nil in unit tests, wired in
  `start()`, pinned by a `ClickProvenanceWiringTests` needle).

The counter is now warp-scoped in fact as well as by design. The
loop-guard model is written as an obligation in
`state-and-layout.md` (§5 row refreshed): an arm refuses its own
re-arm semantically, never via the warp-scoped counter.

Rebased over #688's drain `Policy` restructure; the arm comments
cite the renamed budget the way main now does.

## Related Issue(s)

Fixes #689

## Checklist

  (see AGENTS.md §3)
  required)
  passes
  before merging; it reports, it does not block)

## How I verified

- Full gate post-rebase: 2504 tests / 453 suites green, lint 0,
  site build 0.
- guard-prover: 8 mutations over both arms, the hold, the
  re-fire, the staleness check and the eligibility split — 6 red
  first pass; its 2 inert rows each got a dedicated closure (the
  drain-end wiring needle in `ZOrderSequenceWiringTests` with the
  decrement-before-refire EDGE asserted, and the transient
  focus-regain test), both red-proved inline. The `pointerWarp`
  wiring needle red-proved by deleting the wiring (1 failure,
  behavior suites green).
- Device QA on the signed build is next and will be reported here
  before merge: monocle rapid bar clicks (second restore must
  land), and `mouse.set_follows_focus(true)` with a focus change
  during a restore drain — pointer must follow once the drain
  settles.

## Notes for Reviewer

- Review ran per AGENTS.md §3: code-reviewer + architect-reviewer
  + docs-steward, fixes applied (both majors: the `pointerWarp`
  wiring pin and the warp-test display pinning), re-review waived
  as mechanical closures per the review-change gate.
- Coordinated with #688: merged after it by agreement; the only
  contact points were two comment blocks and the §5 row, all
  resolved in the rebase.


hajiboy95:
Device QA (owner, 2026-08-03) — passed after two rounds of fixes it found, all traced with branch-level logging on device:

- Monocle rapid bar clicks and the cmd-tab-mid-drain warp both pass ("works perfect").
- QA surfaced three linked eaters in the focus-echo routing, fixed in the second commit: a click-made focus report warped (the click pans the row, the settled slot slides out from under the cursor); mouse-made intent held mid-drain re-fired as a spurious late jump (the hold now keeps clickless intent only); and a lazy app's **duplicate** re-report of a raised window threaded both reverts — consume-on-first-echo left it unstamped, and a stale no-echo self-raise entry vetoed the ledger. Stamps now expire by age only, and the veto is age-bounded (`freshSelfRaise`, shared with the self-echo classification).
- guard-prover on the new gates: 6 mutations, 6 red, 0 inert — including both call sites and the shared helper (gutting it reds two independent nets).
- Full gate: 2507 tests / 454 suites, lint 0, site 0.

Issue #678, opened 2026-08-02
📦 Collector: Settings redesign — catalog-driven GUI rebuild


hajiboy95:
### Theme & scope

The Settings redesign — the catalog-driven rebuild of the whole
Settings GUI from the Claude Design project "Settings redesign:
Four approaches" (20 design turns, 19-item decision log, and an
authoritative 281-row placement table). Owned by roadmap #663 as
its own lane beside Wave 1; the working spec, digest and
implementation plan live in the repo's gitignored `plan/
settings-redesign/` (source export kept by the owner).

Design core, in three lines: placement becomes code — a
`SettingKey` enum, `CaseIterable`, catalog built by exhaustive
switch, GUI rendered from the catalog, so an unplaced setting
stops compiling. Home card grid replaces the sidebar; areas
split Easy/Advanced by depth, not scope; the mode never changes
what runs. Settings window tiles like any other window.

Owner sign-offs already recorded (2026-08-02, in chat): the 40
`GUI_REMOVED_2026-08` per-layout bar override rows (incl.
colors) go Lua-only — dependency audit confirmed the Lua chain
is fully independent; design-decisions #68 §3.4 is rescoped,
not repealed, with the superseding entry landing in the Bars
phase.

### Tracked issues

- [x] Phase 1 — `SettingKey` catalog: enum + placement switch +
  the five 4f tests + one-off migration dump reviewed against
  the placement table. Everything else builds on this; no view
  work before it lands. **Shipped in #682** (merged `2f77ff0d`)
  as `Settings/Census/` — 282 rows, gates transcribed from the
  live GreyOut wiring incl. container-level gates, 12 guards
  all red-proofed, `.simple`/`.nerd` mode vocabulary,
  coexistence obligation in `gui.md`.
- [x] Phase 2 — Bars, the proving ground (needs Phase 1): area
  rendered from the catalog, 62→38 rows, per-layout override
  GUI deleted per audit, greying re-check for the Plain
  default, #68 §3.4 superseding entry.
- [x] Phase 3 — COMPLETE (2026-08-04): ~~Home~~ (done, #736 —
  the sidebar-drop; carries the 14c banner, Simple/Power User
  segment, and the Behaviour census conversion) ·
  ~~Layout Defaults~~ (done, #701) · ~~Spaces & Layouts~~
  (done, #727 + #728) · ~~Shortcuts + layers rename~~ ·
[...]

PR #714, opened 2026-08-03
feat(gui): make an app rule read as a sentence, checked against real windows (#678 Phase 3)

hajiboy95:
## Description

#678 Phase 3's **App Rules** lane — spec turn 14a.

**The row is a sentence.** "Spotify opens in media ▾ and floats ▾" replaces an app-name header over two labelled facet columns. The facet values became verb phrases, because a menu inside a sentence has to read as part of one. The **word order is the translator's**: the frame is one key with positional specifiers, split by `SentenceFrame` and emitted in whatever order the translation put them — de and ru genuinely reorder `%3$@` before `%2$@` in the shipped catalogs, which is the design paying for itself inside its own diff.

**A title rule is checked against real windows as you type.** A title fragment is the one part of a rule whose effect you cannot read off the rule — "Info" also catches "Information", and misses "Get Info" because the match is case-sensitive. The editor lists the app's open windows with the verdict beside each, from the engine's own matchers.

**`FirstRunKeymapTests`** pins the seeded ⌃⌥ keymap: a fresh install is usable from the keyboard before the user configures anything. It is a Core seed contract and the only net under it; the #678 day-one Home screen is its most visible consumer, not its owner.

## Related Issue(s)

Part of #678 (Phase 3, turn 14a).

## Checklist

- [x] I understand what this change does and have run it in the app
- [x] New behavior is tested
- [x] `swift build && swift test && ./scripts/lint.sh` passes — 2639 green
- [x] Release build green — CI reports it; no concurrency work in this diff
- [x] PR is focused

## How I verified

Build, 2639 tests, lint and the site build green. Rebased twice onto a moving `main` (#713, #715, #716); the eleven catalogs auto-merged to 885 keys each with zero missing and zero orphans, and the one hand-resolved conflict was the AGENTS.md §5 GUI row, merged as a union of both sides' clauses.

**The localization round landed** — 13 keys × 10 locales, all through `scripts/extract-keys` / `merge-keys` / `drop-key`, no catalog hand-edited.

**`guard-prover` ran five times, `code-reviewer` twice, `architect-reviewer` twice.**

## Notes for Reviewer

**The translation round found a shipped bug no test could have.** The sentence's `HStack` carried `spacing: 6`, so a gap landed between every segment. English hides it — its literals already carry their own spaces (`" opens in "`) — but in ja and ko the literal between two slots begins with a particle (`は`, `에`) that must hug the noun before it, and the gap tore it off. That is the defect undoing `SentenceFrame`'s entire purpose, in exactly the four locales it exists for.

**The guard for it took four attempts, and each prover pass found the previous fix's hole.** Body-scoped → a nested stack walked through. Every-stack-in-the-body → a helper *declared outside* `sentence` and composed in walked through. File-scoped → the file list was hand-written, so a third file walked through. Along the way the pass killed a non-vacuity check I had written that was a **tautology** (`scanned == rowFiles.count`, both sides incremented together, failing branch unreachable) and a **fail-open** (chaining `stripComments` into a literal-blanker: one URL literal leaves an unterminated quote and swallows a real `HStack(spacing: 6)`).

**Two follow-ups are filed rather than fixed here**, both from architect-reviewer, both scoped as prerequisites for the General lane rather than for this branch — the copy they warn about happens the moment a *second* sentence is written, which is the next lane:

- Sentence layout has no owner. A `SentenceFrameView` owning `HStack(spacing: 0)` + `ForEach(frame.segments)` would make the spacing unrepresentable and delete this guard, its allow-list and both helpers outright.
- `SentenceFrame` fuses a general splitter with an App-Rules control vocabulary, so General cannot route to it without adding its nouns to an App Rules enum — and will copy the splitter instead.

Stale-translation defects found in passing and filed as #717: they predate this branch, and the English is the wrong one.


hajiboy95:
Owner eye-confirmed on the branch binary (verified by process path, not by assumption): the sentence row, the verdict list including "floats anyway", and the empty state all read correctly on device.

Two owed items remain before merge — the localization round for the six meaning-changed strings, and a sequential re-review of the fix batch.

Issue #740, opened 2026-08-04
🐛 MonitorReadoutTests fails on any non-English machine (reads the system locale)


hajiboy95:
### KiwiDesk version or commit

`2385fe11` and `1547b8f6` alike — pre-existing, not introduced by either.

### macOS version

macOS 26 (Darwin 25.6.0), system language **German**

### Mac model type

Apple Silicon (arm64)

### Steps to reproduce

1. Set the Mac's preferred language to German (any language with a shipped `Resources/Locales/*.json`).
2. `swift test --filter MonitorReadoutTests`

### Expected vs. actual behavior

**Expected**: the suite passes on any developer machine.

**Actual**: 9 failures, all of the same shape —

```
MonitorReadoutTests.swift:39:9: Expectation failed:
  (empty → "Keine Spaces hier") == "No spaces here"
MonitorReadoutTests.swift:62:9: Expectation failed:
  (one → "1 Space hier") == "1 space here"
```

The suite asserts **English literals** against `MonitorReadout.sentence(...)`, which routes through `L()` → `LocalizationManager.shared`. That manager resolves "System default" from `Locale.preferredLanguages`, so on a German machine it returns German and every literal comparison fails.

Two full-suite runs on this machine failed 9/9. Earlier in the same session two other full runs passed — `LocalizationManager.shared` is a lazily-initialised process-global singleton and swift-testing runs suites concurrently, so the outcome depends on which suite touches it first. That makes it order-dependent as well as machine-dependent.

### Relevant output and logs

```
Test run with 2881 tests in 528 suites failed after 27.931 seconds with 9 issues.
```

CI is unaffected — its runners are English, which is why this has never gone red there.

### Other window managers or AX tools running?

- [x] None that I know of

### Extra context

This blocks `scripts/release.sh` on a non-English machine, since the script runs the verify gate. v0.9.5 was cut with `--skip-verify` on the strength of CI having verified the same commit green (all five checks including Release Build).

`.claude/rules/tests.md` already names both halves of this:

- *"Pin the display, never inherit it … and any default a fixture reasons from"* — the locale is exactly such a default, and this fixture inherits it.
- *"A test that touches process-global state proves itself alone AND in a full run"* — neither a `--filter` run nor a `guard-prover` mutation can observe this class, because both watch the test on its own.

[...]

PR #770, opened 2026-08-07
fix(gui): stop the accent painting bordered button labels

### LINKED ISSUE #759: 🐛 Accent green leaks into button label text across Settings

0.9.5 (426c5cee)

### macOS version

macOS 26.6

### Mac model type

### Steps to reproduce

1. Open Settings.
2. Go to **Gaps & Borders** and look at the "Fit to layout gaps"
   button (`FitGapsAction.swift:36`).
3. Go to **Shortcuts** and look at the "Add application" button
   (`KeybindingAppGroup+AddRow.swift:46`).
4. Go to **App Rules** and look at the "Add app rule" button
   (`AppRulesSection.swift:181`).

**Expected:** the button's label renders in the ordinary label
ink. The accent green marks control *fills* — a toggle track, a
selected segment, a prominent Save.

**Actual:** all three labels (text and SF Symbol) render in the
accent green. `SettingsView.swift:44` applies
`.tint(SettingsTheme.accent)` to the whole tree, and on macOS a
`.buttonStyle(.bordered)` button paints its label in the tint, so
every bordered button in Settings inherits the accent as *text*.

This is a stated rule, not a taste call. From
`.claude/rules/gui.md`:

> **The accent marks control FILLS, never text naming a value.**
> A toggle track, a selected segment, a prominent Save. A
> `.menuStyle(.borderlessButton)` paints its LABEL in the accent,
> so every one of them carries `neutralMenuLabel()`, paired per
> file by the same suite.

`.bordered` is the exact same defect in the exact same shape —
the borderless *menu* half was found and fixed
(`Sources/KiwiDesk/Settings/NeutralMenuLabel.swift`, five call
sites), the button half was not.

### Relevant output and logs

Not applicable — a rendering defect in the Settings GUI, no core
state involved.

Affected call sites: 23 `.buttonStyle(.bordered)` uses across
`Sources/KiwiDesk/Settings/`, including `SettingsFooter.swift`
(×3), `IconPicker.swift` (×4), `NativeSpacesGroup.swift` (×2),
plus the three named in the repro. Full list is a grep for
`buttonStyle(.bordered)` under that directory.

- [x] None that I know of

### Extra context

**Proposed fix — global, not per site.** A `neutralButtonLabel()`
twin of `neutralMenuLabel()`, applied to every `.bordered` button
whose label is prose, plus the same per-file pairing lens in
`SettingsThemeWiringTests` that already pairs the menu version —
so the next `.bordered` button reds instead of shipping green
text.

The pairing lens matters more than the fix: this is the second
time the accent has leaked into text through a control style's
default, and a one-off sweep leaves the third to be found by eye.

Two things the sweep must **not** flatten:

- A genuinely prominent action (`.borderedProminent` Save) is a
  filled control and keeps the accent — as a fill.
- Anything already neutralised deliberately (`KeyRecorderField`'s
  own `.tint(buttonTint)`, `LuaEditorTab`'s `adoptGreen`) is a
  local decision with its own reason; check before overriding.

Part of the #678 Settings-redesign follow-up sweep.

hajiboy95:
## Description

`.buttonStyle(.bordered)` colours its label from the tint, so
once #678 turn 16b tinted the window kiwi every bordered button
in Settings rendered **green text** — "Add app rule", "Add
application", "Fit to layout gaps" and sixteen more.

This is the same defect `neutralMenuLabel()` fixed for
`.menuStyle(.borderlessButton)` on 2026-08-04. Only the menu
half was found then; this is the button half.

Adds `neutralButtonLabel()`, its twin, applied at 21 sites.

**Four bordered buttons are deliberately exempt**, and the
allow-list in `SettingsLabelNeutralityTests` is the one copy of
who may:

- three `role: .destructive` buttons — the system red **is** the
  warning, and neutralising one would paint a delete in the same
  ink as a save;
- `KeyRecorderField`, which resolves its own tint per state. It
  had the bug too: `buttonTint` returned `nil` at rest, which
  inherits the accent, so an idle recorder read green. That arm
  now returns `ink`. The other two stay — red is the flash, and
  the accent while recording is the tint `RecorderButtonChrome`
  *fills* from, so there it marks a surface rather than words.

`.borderedProminent` needs no exemption: it is a filled control,
so its accent is the fill this rule protects, and it never
matches the needle.

## Related Issue(s)

Closes #759

## Checklist

  (see AGENTS.md §3)
  required)
- [x] User-facing changes are documented in `docs/` — n/a: no
  behavior or setting changes, only the ink a label is drawn
  in. No `L()` key moves, so no locale churn either
  (`extract-keys --check`: 971 keys, unchanged).
  passes — see the note on #740 below
  for concurrency / `@Sendable` changes — **not run locally;
  no concurrency or `Sendable` surface is touched, so CI's job
  is the one that counts here**

## How I verified

Ran `.build/debug/KiwiDesk` and walked the Settings tree with
the owner (2026-08-07). Confirmed by eye that the previously
green labels now read in ordinary ink and **match the
dropdowns beside them** — the Spaces layout picker and the
profile selector, which are black because they already carry
`neutralMenuLabel()`. That match was the target.

**The eye-confirm found two the sweep had missed**, both the
same shape: a `Button` with **no `.buttonStyle` at all**. macOS
renders the default style as a bordered push button, so it takes
the tint identically — and the guard's needle, keyed on
`.buttonStyle(.bordered)`, cannot see it. Colours' "Import…"
and Shortcuts' "Import from init.lua…" both shipped green.
Both now name their style explicitly, which is what brings them
under the guard rather than merely recolouring them. The palette
one also took `square.and.arrow.down`, the glyph its Shortcuts
twin already carried — one verb, one symbol, same `L()` call
site.

**Guards, both proved red by mutation** (dropping a
`.neutralButtonLabel()`; turning an exempt button non-bordered),
and the moved menu guard re-proved in its new home:

- `everyBorderedButtonNeutralisesItsLabel` — pairs per file,
  minus declared exemptions;
- `everyBorderedButtonExemptionIsStillUsed` — reds when an
  exemption outlives the button it excuses.

**Suite:** 2899 tests, 9 issues, all `MonitorReadoutTests` —
that is #740 (the suite reads the system locale and fails on any
non-English machine). Confirmed pre-existing by stashing and
running the same filter on clean `main`.

## Notes for Reviewer

**Not yet reviewed — `code-reviewer` and `architect-reviewer`
have not run.** Opening as a draft for that reason; they run
before this comes out of draft.

**One thing no test covers, and the first thing to check:**
does a **disabled** bordered button still grey? `neutralButtonLabel()`
sets `foregroundStyle` explicitly, which can in principle defeat
the disabled dim. "Add app rule" with an empty field is the
quickest case. Not yet eye-confirmed.

**A structural change worth knowing about:**
`SettingsThemeWiringTests` crossed the 350-line hard ceiling, so
both label guards moved into a new `SettingsLabelNeutralityTests`.
The rationale is in that suite's docstring: the defect recurs per
*control style*, so the next style that paints a label from the
tint belongs beside them rather than being found by eye a third
time. Because a move that repoints a test owes a prover run, the
migrated menu guard was re-proved red in its new home.

**Stated residue, deliberately uncovered.** A button with no
`.buttonStyle` stays invisible to the needle. A scan for unstyled
`Button(` matches ~53 sites in `Settings/`, and nearly all are
menu items, context-menu entries and alert buttons whose
containers style their labels — a guard over that set would
demand `.neutralButtonLabel()` at dozens of call sites where it
does nothing or is wrong, which is worse than the gap. The
mitigation is a convention instead, written into the suite
docstring: **an action button names its style.** If a reviewer
sees a cheap way to discriminate those two populations in a
scan, that would close the last hole here.

**Why an allow-list rather than a bare pair** (the menu twin has
no exemptions): unlike menus, there genuinely are bordered
buttons whose label must not be ink. The counts are per file and
per entry, so a file with one legitimate exemption cannot quietly
grow a second.


hajiboy95:
## `architect-reviewer` pass — 1 blocker, 4 major, 3 minor

Recorded rather than actioned: the session ran out of budget.
**Do not take this PR out of draft until the blocker is closed.**
`code-reviewer` ran earlier and its findings are already fixed
(`cdd09187`, `313634f9`).

### Blocker — the "stated residue" is not residue

The suite docstring calls unstyled buttons an accepted gap. It is
not: **nine text-labelled, in-window instances of the shipped
defect remain**, so this PR closes the *guarded* population while
the bug stays on screen — and the §5 obligation it introduces
("a control style that colours its label from the tint owes a
neutralising modifier at every call site") ships already
violated.

- `Sections/GeneralSection.swift:190` — "Reveal"
- `Sections/GeneralSection+Reset.swift:30`, `:74` — "Discard
  Saved Window Arrangement", "Reset All Settings…"
- `Sections/AppRuleTitledEditor.swift:37` — "Add"
- `Components/General/LoginItemCard.swift:225` — "Open Login
  Items"
- `Components/Common/PermissionPausedBanner.swift:39` — "Enable
  Accessibility…", green on the warning surface
- `Sections/ProfilesSection+Rename.swift:47` — "Rename"
- `Components/Lua/LuaEditorTab.swift:100` — "Back to visual
  editor"
- `Components/Icons/IconPicker.swift:215` — "Use … as text", in
  the same file whose four bordered buttons the guard now pairs

**And the proposed guard shape is better than what I argued
for.** I rejected guarding unstyled buttons because ~53 matches
are mostly menu/alert items. The reviewer's answer is to guard
the *convention the rule already states* instead of the
neutralisation: require an explicit `.buttonStyle` on any
`Button` not lexically inside a `Menu` / `contextMenu` /
`confirmationDialog` / `alert` closure. That is tractable, and it
demands nothing of the ~40 menu-item sites.

### Major

1. **`owed = styled - exempt + other` is position-blind**
   (`SettingsLabelNeutralityTests.swift:193`). In the one file
   where both registers fire — `SpacesSection+Overrides.swift`,
   `styled=1 exempt=1 other=1 → owed=1` — moving the single
   `.neutralButtonLabel()` off the `.borderless` breadcrumb
   (`:116`) onto the destructive Delete (`:198`) keeps the count
   at 1, kills the system red the exemption exists to protect,
   and `borderedExemptionsAreLive` still passes. The suite makes
   exactly this argument for `borderedExempt`'s `needle` and then
   omits a needle from the second register.
2. **`neutralOnOtherStyles` has no precedent in this repo's
   allow-list idiom** (`:154`). `ActivationPolicySeamTests` and
   `SettingsColorSurfaceTests` assert membership or whole-map
   equality; neither carries an inflation offset. It exists only
   because one modifier name now serves two styles — the
   per-style names were what made per-file pairing attributable,
   and this PR spent that property. Fix: a named modifier for the
   `.borderless`-with-text case, or pair style + neutralisation
   **by construction** in one `settingsActionButton()` helper —
   the repo's seal-over-guard shape — which deletes the register
   outright.
3. **The recorder ships half the fix**
   (`KeyRecorderField.swift:166`): `.tint(buttonTint)` with no
   `.foregroundStyle`, while both modifiers argue the pair is
   required because an AppKit-backed label follows the tint "and
   neither is redundant on every macOS". Either the recorder owes
   `.foregroundStyle(buttonTint)`, or that claim is wrong and
   both modifiers are over-specified. The two cannot both stand.
4. **A third tint-reading style is unrecorded**: `.buttonStyle(.link)`
   with TEXT labels at `Components/Colors/PaletteShelf.swift:140`
   ("Pair with Glow") and
   `Components/Keybindings/Recorder/KeyRecorderRejectionRow.swift:53`
   ("Steal", "Go to"), all rendering kiwi. A link taking the
   accent is probably right — but the register has to *say so*,
   the way it does for `.borderless`, or the next author reads a
   two-style list as the considered set. That is the exact
   failure that produced this PR.

### Minor

- `KeyRecorderChrome.swift:13` — "(the tint alone only recolors
  the border)" narrates a recording-state recolour this branch
  removed.
- `.claude/rules/gui.md:476` — "two are held there today" is a
  hand-kept count of another file's contents with nothing
  deriving it, stale the moment `.link` joins. The enumeration
  carries the meaning; drop the number.
- `PaletteShelf.swift:78` — the new SF Symbol on "Import…" is a
  visible design change riding in a colour fix, outside the
  stated scope and not covered by the owed greying eye-check.

### Settled, needs no further work

Scoping the root `.tint(SettingsTheme.accent)` down instead of
neutralising per style is **the wrong trade** — ~137 legitimate
accent consumers in `Settings/` (25 `Toggle`, 48 `Picker`, 28
`SegmentedPicker`, 24 `Slider`, 12 `.borderedProminent`) against
~24 neutralisations, and a forgotten scoped tint yields a
system-blue fill, i.e. the `Color.accentColor` semantics
`SettingsThemeWiringTests` retired. One root wiring point with a
guarded opt-out list is the right shape, and the twin modifier
files are fine under §2.4 — it is the twin *names* that are
load-bearing, and they are undocumented.

### Still owed besides the above

The disabled-greying eye-check ("Add app rule" with an empty
field) — `neutralButtonLabel()` sets `foregroundStyle`
explicitly, which can defeat the disabled dim. Uncovered by any
test.

hajiboy95:
## Architect findings addressed — `bcd91e87`

**Blocker closed.** Nine unstyled text-labelled buttons now name
a style and carry `neutralButtonLabel()`: "Reveal", both Reset
actions, "Add", "Open Login Items", "Enable Accessibility…",
"Rename", "Back to visual editor", "Use … as text". With the two
Imports fixed earlier that is eleven, and the suite docstring now
says so instead of calling them residue.

**Major 1 — position-blind arithmetic: fixed.** The declared
style in `neutralOnOtherStyles` is asserted **adjacent** to the
modifier. Proved red by the exact mutation named in the review —
moving the single modifier off the `.borderless` breadcrumb onto
the destructive Delete, which previously kept every count
balanced.

**Major 3 — half a fix in the recorder: fixed.**
`KeyRecorderField` now sets `.foregroundStyle(buttonTint)` beside
`.tint`. `KeyRecorderChrome`'s docstring no longer argues from a
border recolour this branch removed and states it is the whole
recording signal.

**Major 4 — `.buttonStyle(.link)`: recorded**, left accent
deliberately, with the reason written down.

**Minors** — gui.md's hand-kept count dropped; the stale chrome
parenthetical rewritten.

### Still open, deliberately

**Major 2 — `neutralOnOtherStyles` has no precedent in the
allow-list idiom.** The suggested fix is a `settingsActionButton()`
helper pairing style + neutralisation *by construction*, which
would delete the register outright. Not attempted: it touches
every call site this branch just changed, so it wants its own
change rather than riding a fix batch at the end of a session.

**The convention guard.** "An action button names its style" is
still unguarded. The reviewer's shape — require an explicit
`.buttonStyle` on any `Button` not lexically inside a `Menu` /
`contextMenu` / `confirmationDialog` / `alert` closure — is
better than what I built and is recorded in the suite docstring
as not-yet-built. That guard is what would have caught all eleven.

**The disabled-greying eye-check**, unchanged.

`swift build`, `scripts/lint.sh` and the three affected suites
green (10 tests). Still draft.

hajiboy95:
**Disabled-greying eye-check: PASSES** (owner, 2026-08-07).

"Add app rule" with no app chosen still renders grey, so the explicit `foregroundStyle(SettingsTheme.ink)` in `neutralButtonLabel()` does **not** defeat the disabled dim — the one risk the modifier introduced and the only item on this PR that no test could cover.

That clears the second of the three open items. Remaining before undraft:

1. **The convention guard** — "an action button names its style", unguarded. The architect's shape: require an explicit `.buttonStyle` on any `Button` not lexically inside a `Menu` / `contextMenu` / `confirmationDialog` / `alert` closure.
2. **`settingsActionButton()`** — recommend descoping to its own issue rather than blocking this PR. It is a refactor of a working fix; the guard is the protection.

hajiboy95:
## Convention guard landed — `07f5aeba`

`SettingsButtonStyleConventionTests`: every `Button` in `Settings/` names a style, except one lexically inside a `Menu` / `.contextMenu` / `.confirmationDialog` / `.alert` closure. Brace-tracking lexer over comment-stripped source. Red-proofed both ways — stripping the style from `deleteLayer` fails it, a styleless button nested in a `Menu` is correctly ignored. Remaining violations were fixed by naming a style, not by exempting them.

Authored by Gemini; lint-wrapped, verified and committed here (it arrived with lines up to 141 chars against the 79 limit, so `scripts/lint.sh` was red on delivery).

**All three open items are now closed** — blocker, majors, greying eye-check, convention guard. Build, lint and the affected suites green.

### One thing to weigh before merging

The guard is **count-based**: `unexcludedButtons == (styles - extraStyles) + exempt`. It enforces the convention, but a count cannot say WHICH button carries the style — the same position-blindness the architect flagged on `neutralOnOtherStyles`. To balance, it adds two more registers: `unstyledExempt` (sites the lexer cannot reach — a button rendered from a separate `@ViewBuilder` property used inside a context menu is invisible to lexical containment) and `stylesOnNonButtons` (`.buttonStyle` on a `Picker`).

So this family now keeps **five** hand-written registers for one invariant. That is not a reason to hold the PR — the guard is the only thing standing between a new action button and the defect — but it sharpens #771: pairing style and neutralisation *by construction* would delete four of the five.

Worth noting the two register kinds are different in nature and share one map shape: a genuine exception (`role: .destructive`) and a scanner blind spot are not the same thing, and mixing them means a real violation can hide behind "the lexer could not see it". #771 is where that gets resolved.

Ready to undraft.

Issue #488, opened 2026-07-24
fix(commands): focus left/right navigation misbehaves on bsp and track


hajiboy95:
Reported during #482/#483 device QA (2026-07-24): directional `focus` (left/right) "seems somehow bugged" on **bsp** and **track** layouts. No precise repro captured yet — needs a fresh session that starts by pinning down the symptom (wrong neighbor picked? no-op? focus jumps across displays?).

## Context for the digging session
- Cross-layout navigation must respect each layout's model (AGENTS.md §5): bsp is *geometric* (neighbor search over calculated slots), track is *geometric with overflow piles* (`Navigation.pileMates`, #172). The authoritative map is the "Layout navigation & overflow models" table in `docs/design-decisions.md`.
- Recent changes touching focus/raise paths worth suspecting first: #463 settle re-assert, #446 display-follow, #444/#449 display-resolved slot math (geometric neighbor search now resolves per-display frames — a wrong display resolution would corrupt bsp/track candidate sets), and the #482/#483 move-focus latch (should be inert here — `focus` navigates within the active space — but verify).
- Repro via CLI (`KiwiDesk focus left` etc.) so the `CommandResponse` is visible; the hotkey path swallows `.fail`, and the #292 preflight now logs denials (`preflight (#292): denied …`).

## First steps
1. Get an exact symptom description + window arrangement from the owner (which space mode, how many windows, single or multi display).
2. Check whether the wrong-neighbor choice is geometric (slot math) or a #292 denial / dropped raise masquerading as a no-op.

hajiboy95:
Owner repro detail (2026-07-24):

- **track**: 4 tracks side by side (horizontal, left→right); directional focus could not reach every track — some tracks unreachable by wandering left/right.
- **bsp**: 4 windows — 1 left half; right half split into 1 top-right and 2 bottom. Focus moved fine between the leftmost window and the two bottom windows, but could never reach the top-right window.

Both smell like the geometric neighbor candidate search (shared by bsp/track) discarding valid slots — the bsp case (can't go *up* to top-right from bottom-right) suggests the vertical axis candidate set or the tie-breaking is wrong, not a raise/denial issue.

PR #885, opened 2026-08-18
feat(layout): monocle hide_style — stack stays default, park opts in (#881)

### LINKED ISSUE #881: 💡 Monocle: park unfocused members at the stash corner so transparent apps don't reveal the stack, with an instant switch
### Use case

Monocle stacks every window at the same full-screen frame, and
only the focused one is supposed to show. But an app with a
transparent or blurred background (a terminal with opacity, any
vibrancy surface) shows the windows STACKED BEHIND it through
its own body — the layout's illusion of "one window" breaks on
exactly the apps that are popular to run in monocle.
Owner-observed during #880's live QA round (2026-08-18).

### Proposed solution

Park the unfocused monocle members at the hidden-window corner
instead of stacking them behind — the same corner and geometry
the space stash already uses — and make the focus switch
INSTANT (no slide animation) so it keeps today's raise-only
feel. Investigated with the mechanism in hand; the pieces
exist:

- `TilingEngine.stashFrame(_:in:corner:)` and
  `optimalHideCorner` are `nonisolated static` and
  size-preserving, and `ScreenNeighbors` is in `LayoutContext`
  since #878 — so `MonocleLayout` itself could stay a pure
  function and emit the focused window at the full frame plus
  every other member at its park frame. No engine surgery.
- Monocle would need a retile on focus change (scrolling
  already retiles on focus), passing `animated: false` for
  this transition so both moves (park the old, place the new)
  snap through the `applyInstant` path — the "feels like it
  currently does" requirement.
- `restoreMonocleZOrder`'s raise stays for the placed window;
  its stacking concern mostly dissolves (nothing overlaps).

Decisions the ruling has to make, found by the investigation
rather than guessed:

1. **The corner slivers.** A parked window keeps a ~1×48 pt
   visible tab (the WindowServer visibility floor — the stash's
   own accepted geometry). N monocle members = N slivers piled
   in the corner. Today's stack hides everything perfectly
   BEHIND the focused window; the proposal trades
   visible-through-transparency for slivers-in-the-corner. The
   space stash already made this trade and it was accepted —
   but there the slivers belong to another space; here they sit
   on the ACTIVE one.
2. **Default change vs option.** Changing monocle's behavior
   outright alters a frozen surface; an option
   (`monocle.hide_style = stack | park`) adds a configuration
   axis, which the 1.0 guiding rule defers. Either way this is
   post-1.0 material unless re-ruled.
3. **Mission Control / app-switcher thumbnails** show parked
   windows at the corner (already true for stashed spaces —
   same precedent).
4. Interactions checked and unalarming: tiled-sticky travelers
   (park like members), native-fullscreen members (already
   outside the tiled derivations, #670), the #677 bound
   learner (parked windows leave the layout loop, no asks
   recorded), `looksStashed`/#412 (corner frames already
   classify as stash echoes — that classifier's purpose), the
   create fold (newcomer places focused, siblings park).

### Alternatives considered

- Truly hiding other apps' windows: no public API (only
  minimize, which changes Dock state and is slow — wrong).
- Moving them to another native Space: private, SIP-gated
  surface (the #140 collector's class) — not shippable.
- Ordering them lower in z: they are already behind; a
  transparent body shows whatever is behind it regardless.
  Parking is the only viable shape found.

### Extra context

Surfaced during #880's device QA (the #677/#618 branch), where
monocle got the learned-bound centering — the transparency
artifact becomes more noticeable once the layout otherwise
behaves. Filed unmilestoned by the 1.0 guiding rule (behavior
retune / possible new axis, not a defect against the frozen
surface); Priority Low per the roadmap ladder, Effort Medium (a
lane: one layout, its engine seam, tests, review).

hajiboy95:
## Description

Monocle gains `hide_style` — `stack` (default, byte-for-byte
today's behavior) or `park` (#881): unfocused members park at
the stash corner so a transparent/blurred body, or the side
gaps around a #880-centered width-bound window, cannot reveal
the stack; the focus switch snaps through the instant path so
it keeps the raise-only feel.

- Lua `monocle.set_hide_style("stack"|"park")`, JSON
  `layout.monocle.hide_style` (SettingsCodingTests pins it);
  per-space override deliberately not added (exempted with the
  reason in MonocleOverrideTests).
- `MonocleLayout` stays a pure function: the shown member is
  the context anchor (held across a float focus by the engine's
  `monocleShownMembers`, the `stashedFrames` precedent), every
  other member parks at its own bound-consumed size, corner
  from the shared `optimalHideCorner(neighbors:)` copy, anchored
  past the monocle bar's strip.
- GUI: "Hidden windows" picker on the Monocle card (census row,
  readout arm, interpolated help popover naming both cases),
  schematic fan collapses under park with an iconic corner pile
  at `.panel`, captions/ax switch with the control.
- One translation round: 7 keys × 10 catalogs (drafter + a
  separate read-only audit; fr re-settled its park verb on
  *garer* after a collision with the Scrolling row noun, ru
  fixed to its informal register).
- Docs: lua-reference, cli.md (also corrects the stale monocle
  `wrap_focus` default claim), user-guide, design-decisions
  entry arguing the option-not-default ruling.

## Related Issue(s)

Closes #881

## Checklist

  (see AGENTS.md §3)
  required)
  passes

## How I verified

Full gate per commit: `swift build`, `swift test` (3568 tests,
669 suites, green), `scripts/lint.sh`, plus the site build for
the docs pages. guard-prover (worktree-isolated) red-proofed
all 15 new/changed assertions via sub-diff mutations — none
vacuous. Review rounds: code-reviewer + architect-reviewer +
docs-steward + localization-auditor, findings fixed (park
anchored past the bar strip, shown-member hold across a float
focus, fr/ru redrafts) and re-reviewed clean. The signed app
was rebuilt (`scripts/build-app.sh`) and relaunched from
`.build/app/KiwiDesk.app`.

**Device QA: done by the owner (2026-08-18)** — wrap arms,
instant switch and the ring-at-switch confirmed after the QA
round's fixes; own-window focus observations filed as #887.
The original checklist: under `park` — translucent terminal shows the desktop,
a centered width-bound window shows clean gaps, the switch
feels instant, the corner reads fine on a single-Space setup;
under `stack` — byte-for-byte today's behavior. That eye-round
is the owner's.

## Notes for Reviewer

- The z-order arm stays armed under park on purpose — the
  rationale is written at the arm (`KiwiCore+FocusRaise`): the
  stacking half dissolves, the bar-click reliability half is
  style-independent.
- The Mission Control claim in the docs rides the issue's own
  investigation (same mechanism as stashed Spaces); a
  ten-second on-device glance confirms it.
- Unmilestoned post-1.0 work by the 1.0 guiding rule (a new
  config axis) — the 1.0 count does not move.


hajiboy95:
Owner device QA round (2026-08-18), both findings fixed on-branch and re-proved:

1. **`wrap_focus` looked inert under `park`** — the unwrapped cycle's dead end fell through to the geometric focus tier, whose candidates now include the parked slivers' real corner frames; focusing a parked member *shows* it, indistinguishable from wrapping. The parked members leave the tiled tier (`MonocleLayout.shownMember` is the one copy of the pick, shared with the layout; the shown member stays so the #488 float→tile hop still lands; `stack` untouched). Both wrap arms are observable again.
2. **The focus ring trailed a fast park switch** — the instant path has no animation ticks and the steady overlay sync read only the echo-fed state frame. `FollowSource.syncFrame` gains a `commanded` input (the applier's just-issued instant target, retired by the first self-echo), so the ring and the sticky mark move at the switch.

Commits 5b033b64 + 4dcfd837; guard-prover red-proofed all seven new assertions (one test was inert against its own filter on the first cut and was re-fixtured to discriminate); full suite 3574 green. The relaunched app carries both — re-check the wrap toggle and the fast-switch ring under park.

Issue #902, opened 2026-08-21
💡 Config files carry no format stamp, so a migration can never be removed


hajiboy95:
### Use case

I want to be able to **delete a config migration** one day.

`ConfigMigration` (added on `feat/app-bar-label-style`) rewrites
`app_bar.content` in profiles written by v0.9.7 or earlier, where
the value was `name` / `icon_and_name`. It runs on every
`ProfileManager.read`, forever, because nothing can tell it that
the last config carrying the retired spelling is gone.

Neither `Profile` nor `GuiConfig` carries a format version. So
there is no answer to "has this file been migrated?" — only "does
it currently parse?". That is not the same question, and the
difference is what makes removal unsafe:

- A user upgrading v0.9.7 → 1.1 in one step never runs whichever
  build did the rewrite. If the migration was deleted in 1.0
  because "everyone has updated by now", their profiles fail to
  decode, `allProfiles()` skips them, and the profiles disappear
  from the list.
- That failure is silent on the developer's machine by
  construction: our own configs were migrated by the build we
  were running while we wrote the code.

`SetupBundle` already has the mechanism (`currentFormat = 1`,
checked on import). The two files read on **every launch** do not.

### Proposed solution

Give `Profile` and `GuiConfig` a `format` integer, following
`SetupBundle.currentFormat`:

- **Write** the current format on every save.
- **Read** it before decoding, and route by it: run the
  migrations for formats below current, then decode.
- **Absent** means format 0 — a file written before this issue
  landed — which is exactly the set the current migration
  targets.
- **Above** current means a file from a newer build: refuse it
  with a clear message rather than half-decoding it. (`SetupBundle`
  already does this: `format <= SetupBundle.currentFormat`.)

Then removing a migration becomes a decision with a floor:
"we migrate from format N-1; a file below that is refused with a
message naming the version that can still read it" — instead of
a guess that fails on someone else's machine.

Two smaller pieces that fall out, worth doing in the same change:

- The migration should key off the format, not off pattern-matching
  the payload (today it walks the JSON looking for a retired
  `content` value on every read of every profile).
- `ConfigMigration`'s docstring currently says removal is a guess
  *until this exists*. It should be updated to name the floor.

### Alternatives considered

**Leave it as-is and never remove the migration.** Works, and
costs a JSON walk per profile read plus a growing pile of
crossings that no one can ever prove is dead. This is the status
quo, and it is what the issue exists to end.

[...]

PR #945, opened 2026-08-22
fix(config): stamp the migration crossing and bring palettes into the census

### LINKED ISSUE #938: 🐛 ConfigMigration write-back never stamps the format, so the crossing never ends

0.9.9 (main @ 65a658da)

### macOS version

macOS 26.6 (Darwin 25.6.0)

### Mac model type

Apple Silicon (dev machine)

### Steps to reproduce

1. Have a pre-#909 config file carrying a retired value the
   migration rewrites (the `content` rename).
2. Launch: `GuiConfigStore.load` / `ProfileManager.read(name:)`
   apply `ConfigMigration.migrated(_:)` and write the file back.
3. Inspect the file: the migrated bytes carry no `format` key.

Expected: the one-shot migration ENDS — the rewritten file is
stamped with the current `format`, so `needsMigration` goes
false and a future floor advance recognizes it as crossed.
Actual: `migrated(_:)` rewrites values but never stamps
`format`, and both write-back sites persist the unstamped bytes.
The file reads as format 0 forever unless a later user edit
re-encodes it: `needsMigration` stays true on every load (the
skip fast path never engages), and when a future floor advance
deletes this migration, an already-migrated file is
indistinguishable from one that never crossed and is refused as
below-floor — defeating the #909 floor's own contract
(profiles.md: the crossing must END).

### Relevant output and logs

Found by the v0.9.8..main line sweep (2026-08-22),
`Sources/KiwiDeskCore/Config/ConfigMigration.swift:102` as of
that range. Related from the same sweep: `needsMigration`'s
shape sniff hand-mirrors marker keys ("writtenBy",
"monitor_sets") with no tie to the owning CodingKeys — renaming
`monitor_sets` (the rename class this file serves) silently
re-routes profiles onto `GuiConfig.currentFormat`'s floor; give
it a parity pin when touching this.

None that I know of

### Extra context

Fix: the migration write-back also writes the current `format`
(surgical key insert, same promise as the value edit), plus a
test that a migrated file round-trips as
`needsMigration == false`.

### LINKED ISSUE #939: 🐛 palettes.json sits outside the migration census and carries no format

0.9.9 (main @ 65a658da)

### macOS version

macOS 26.6 (Darwin 25.6.0)

### Mac model type

Apple Silicon (dev machine)

### Steps to reproduce

Structural: `PaletteStore.userPalettes()` decodes
`[ColorPalette].self` from `palettes.json` — a bundle-carried,
backup-travelling file shape — while
`ConfigMigrationRoutingTests`' `fileShapes` census lists only
`Profile`/`SetupBundle`/`GuiConfig`. `palettes.json` is also
array-rooted with no `format` key, so
`ConfigMigration.needsMigration`'s `[String: Any]` cast returns
false even if it were routed.

Expected: every reader of a migratable file shape is in the
census (routed, or exempt with a reason), per profiles.md — a
`ColorPalette` field rename then reaches the standing
`palettes.json`, not only the bundle walk.
Actual: the first `ColorPalette` rename will migrate backups but
silently skip `palettes.json`, with the census green.

### Relevant output and logs

Found by the v0.9.8..main architect sweep (2026-08-22).
profiles.md's bump obligation already names a ColorPalette
rename as the breaking case.

None that I know of

### Extra context

Fix: add `ColorPalette` to `fileShapes` with `PaletteStore` in
the census, and rule the palettes format question (array root vs
a wrapped `{format, palettes}` shape) BEFORE the first field
rename.

hajiboy95:
## Description

1. **Stamp format on migration (#938)**: `ConfigMigration.migrated(_:)` now stamps the current format integer (`Profile.currentFormat`, `GuiConfig.currentFormat`, or `ColorPalette.currentFormat`) when migrations apply. This ensures migrated files decode with current format versions on subsequent loads and do not continually report `needsMigration == true`. Shape markers (`writtenBy`, `monitor_sets`, `palettes`) are tied to owning `CodingKeys`.
2. **Bring `palettes.json` into migration census (#939)**: Wrapped `palettes.json` inside a versioned `PaletteDocument` container (`{"format": N, "palettes": [...]}`) with `ColorPalette.currentFormat = 1`. Added a one-shot migration step `migratingLegacyPalettesArray` to lift legacy bare arrays into the wrapped format, routed `PaletteStore.userPalettes()` through `ConfigMigration.migrated`, and added `PaletteStore` to the `ConfigMigrationRoutingTests` census.

**Backup path check**: `SetupBundle` embeds decoded `[ColorPalette]` objects rather than raw file bytes of `palettes.json`. Therefore, `SetupBundle`'s schema and restore path are unaffected by `palettes.json`'s outer envelope wrapper.

## Related Issue(s)

Closes #938. Closes #939.

## Checklist

- [x] `swift build && swift test && ./scripts/lint.sh` passes
- [x] Release build green — CI's `Release Build` job (read it before merging; it reports, it does not block), or locally for concurrency / `@Sendable` changes

## How I verified

Ran full verification suite: `swift build && swift test && ./scripts/lint.sh` (3,780 tests passed, Lint OK).

### Mutation proofs

| Invariant under test | Mutation | Result |
|---|---|---|
| Migration write-back stamps format (#938) | Remove `stamped` format call from `ConfigMigration.migrated` | `ConfigMigrationTests.migratedProfileStampedWithCurrentFormat` fails |
| Bare-array palettes lift (#939) | Remove `migratingLegacyPalettesArray` step from `ConfigMigration.steps` | `PaletteStoreTests.legacyPalettesFileMigrates` & `ConfigMigrationTests.legacyPalettesArrayMigrates` fail |
| Census coverage for PaletteStore (#939) | Remove `Appearance/PaletteStore.swift` from `ConfigMigrationRoutingTests.readers` | `ConfigMigrationRoutingTests.readersRouteThroughMigration` fails |
| Newer palette format rejection (#939) | Remove `decodedFormat <= ColorPalette.currentFormat` check from `PaletteDocument.init(from:)` | `PaletteStoreTests.newerPaletteFormatReturnsEmpty` & `ConfigMigrationTests.newerPaletteFormatIsRefused` fail |

## Notes for Reviewer

Changes are restricted to `Sources/KiwiDeskCore/Config`, `Sources/KiwiDeskCore/Appearance`, `Sources/KiwiDeskCore/Profiles`, and test targets.


hajiboy95:
**Review round complete** (code-reviewer + architect-reviewer +
guard-prover, 9 mutations run). Solid core — the stamp seam
placement, unknown-key preservation, the lift being a real
migration step, and the SetupBundle/restore path are all right,
and 8 of 9 mutations red-proofed. **Not mergeable yet**: 2
blockers, 3 majors. Claude is applying the fixes on this branch.

**Blockers**
1. `ConfigMigration.migrated` — `guard changed else { return nil }`
   never stamps a stale-format file that no step rewrites (an
   unversioned file with no retired value): `needsMigration` stays
   true forever and a future floor advance (#902) refuses a valid
   file. Fix: once `needsMigration` was true, stamp regardless;
   nil only when the bytes are unchanged. Owes the stamp-only test.
2. `PaletteStore.userPalettes` swallows the newer-format refusal
   (`try?` → `[]`), indistinguishable from an empty library — a
   downgraded build's next `save` clobbers the newer build's whole
   library, and the backup export silently exports empty.
   `newerPaletteFormatReturnsEmpty` pins the clobber-enabling
   behavior as correct. Fix: refusal ≠ empty; mutating paths and
   the bundle export refuse; retune the test.

**Majors**
3. `savedPalettesAreWrapped`'s `doc.format == currentFormat` is
   vacuous — the decoder normalizes `format` to current for every
   accepted input, so `write()` stamping `format: 0` leaves ALL
   suites green while reintroducing the endless-remigration
   symptom on the save path (proved by mutation). Fix: keep the
   decoded format, assert the raw saved JSON /
   `!needsMigration(savedBytes)`.
4. `targetFormat`'s `"palettes"`/`"writtenBy"` markers are
   literals and `shapeMarkersMatchCodingKeys` restates them in
   the test (parity costume — breaking the marker leaves all
   suites green, masked today only by all formats being 1). Fix:
   derive in place (`PaletteDocument.CodingKeys`, a marker owned
   by `SetupBundle`), replace the test with one that encodes each
   real shape and routes it.
5. `importPalette`/`export` read/write bare `ColorPalette`
   sidecar bytes outside the census — a future palette key
   migration would silently strip exported files. Rule it
   deliberately (exempt-with-reason recorded in the census; the
   bare root has no marker so routing would misroute it today).

**Minors**: `currentFormat` homed on `ColorPalette` instead of
the file root `PaletteDocument`; the new double-bump obligation
(palettes floor + `SetupBundle.currentFormat`) unrecorded in
profiles.md/§5; the any-top-level-array=palettes assumption
undocumented; `ConfigMigration.swift` (325) and
`PaletteStore.swift` (274) past the 250 target; the census pair's
joint-absence blind spot worth a docstring line.

hajiboy95:
Re-review verified all round-1 findings fixed (stamp-on-every-stale-crossing, loud palette refusal, non-vacuous format asserts, derived markers, sidecar ruling recorded). Its one new minor — the palette refusal narrated as a disk-write failure — landed as `SetupBundleError.unreadablePalettes` with its own localized message in all eleven catalogs, proven red by deleting the pre-check. Merged.

Issue #94, opened 2026-07-09
feat(settings): contextual help (?) affordance for settings fields


hajiboy95:
### Use case

Some settings need a short explanation the user can reach without cluttering the pane. Example: **Layout ▸ BSP ▸ Split strategy** offers two named options ("Shortest side" / "Alternating") whose names don't explain what they do. We want a consistent, macOS-native way to surface a sentence or two of help per field.

### Proposed mechanism

A small circled-`?` (`questionmark.circle`) button at the **trailing end of the row**, opening a `.popover` on click. Chosen over the alternatives because:

- **Hover-only `.help()` tooltips** are fine for one-line hints on ambiguous icon buttons, but not for paragraph explanations: OS-styled/single-line-biased, keyboard-inaccessible, no VoiceOver focus target, and invisible to trackpad users who never rest the pointer.
- **A click popover** matches what Apple's System Settings actually does: a visible, discoverable glyph; a real focusable `Button` (keyboard + VoiceOver accessible); dismissible and re-readable; and closed-by-default so it adds minimal visual noise.
- Keep an always-visible section-level `caption` (existing `SettingsSection` pattern) for section framing only — do **not** add per-field always-visible captions (too heavy on dense tabs).

Also attach `.help(...)` to the same button as a cheap one-line hover fallback, and set `accessibilityLabel`/`accessibilityHint` so VoiceOver announces the affordance and content.

### Placement

- **Per-field help:** the `?` sits **trailing the row, after the control** — never in/before the label column, which would break the shared `settingsLabelColumn` alignment that `DropdownRow`/`SegmentedPicker`/`StepperRow`/sliders rely on. Trailing placement is layout-neutral.
- **Per-option help:** don't hang a `?` off each segment/option (wrecks `SegmentedPicker` geometry; dropdown options aren't on screen). When a field has only 2–3 options, fold per-option text into the single field-level popover (option name in bold, one line each). A dedicated per-option affordance is a rare case to design only if the audit finds a real example.

### Reusable component (sketch)

A `HelpButton(explanation:)` colocated with the shared row primitives in `SettingsRows.swift`:
- `Button` with `Image(systemName: "questionmark.circle")`, `.plain`, `.secondary`, `.controlSize(.small)`.
- `.popover(isPresented:)` showing a fixed-max-width `Text(explanation)`.
- `.help(explanation)` hover fallback + accessibility label/hint.
[...]

PR #950, opened 2026-08-23
feat(gui): add ⌃. keyboard chord for contextual menus on focused rows (#845)

### LINKED ISSUE #845: 🐛 Context-menu-only actions are unreachable from a plain keyboard

`2b764aa8`. Long-standing — every context-menu-only action has
been in this position since it shipped.

### macOS version

macOS 26.6.1

### Mac model type

MacBook Pro (Apple M1 Max)

### Steps to reproduce

Owner question during #842's device QA, 2026-08-12: *"how can I
do right-click with the keyboard?"*

1. Turn System Settings ▸ Keyboard ▸ Keyboard navigation ON and
   put VoiceOver away — an ordinary keyboard user.
2. Open Settings ▸ Advanced Colours and Tab to a saved palette
   tile.
3. Try to reach **Rename…**, **Export…** or **Delete**.

**Expected:** every action the app offers is reachable without a
pointer. `.claude/rules/gui.md` ▸ The keyboard path opens by
saying "accessible with VoiceOver" and "usable without a mouse"
are two claims, and this is the case where the second one fails.

**Actual:** those three actions live behind `.contextMenu`
alone. macOS has no default key that opens a focused control's
contextual menu, so on a plain keyboard they are unreachable —
the tile takes focus and Space applies the palette, and that is
all a Tab-only user can do with it.

The mirror exists and works, but only for one audience: every
`.contextMenu` in this tree is also offered as
`.accessibilityActions` from the same builder
(`KeyboardActionParityTests`), which VoiceOver reaches via
`VO`+`Shift`+`M`. Outside VoiceOver, the paths are Mouse Keys
(`Ctrl`+`5` on a numpad) or nothing.

### Relevant output and logs

Not applicable — an affordance gap, no runtime state.

- `Sources/KiwiDesk/Settings/Components/Colors/PaletteShelf.swift`
  — `userMenu` behind `.contextMenu` + `.accessibilityActions`.
- The same shape on the draggable rows the ruling below was made
  about (spaces, profiles, layers).
- `.claude/rules/gui.md` ▸ The keyboard path records the residue
  as accepted: "this reaches VoiceOver, not a Tab-only keyboard".

- [x] None that I know of

### Extra context

**This is a ruling to re-make, not a defect to fix quietly.** A
visible `⋯` trigger per row was proposed and rejected twice as
clutter (owner 2026-08-04, upheld 2026-08-11 against turn 20a's
ask), and a whole-chip `Menu` was rejected because it eats the
`.draggable` it would sit on. Both rejections assumed
`.accessibilityActions` was an adequate second path. This issue
is the argument they did not weigh: that path requires VoiceOver
to be running, so for a keyboard user with no screen reader the
actions do not exist.

Options, so the decision is made once and applied everywhere
rather than per surface:

1. **A `⋯` on hover AND on focus.** Keeps the resting row clean
   — which is what the two rejections were protecting — while
   giving the keyboard a target, since a focused row is exactly
   when the trigger is needed. `gui.md` bans a hover-revealed
   control inside a pill, so this needs a shape ruling per
   surface, not a blanket yes.
2. **A key on the focused row** (`⌘`+`.` or similar) opening the
   same menu. No chrome at all; the cost is discoverability —
   an invisible key is only found in documentation.
3. **Accept it, and say so where a user reads it.** Legitimate,
   but then `docs/user-guide.md` ▸ Using Settings from the
   Keyboard should state that these actions need VoiceOver or a
   pointer, rather than the tree carrying the residue silently.
   `docs/accepted-limitations.md` is where that row would live.

Whichever wins binds every `.contextMenu` in the tree, since the
parity guard already treats them as one family.

hajiboy95:
## Description

Implement owner ruling for #845 (Option 2 — a keyboard chord on the focused row).

Actions behind `.contextMenu` alone were previously unreachable for plain keyboard users navigating by Tab (with macOS full keyboard navigation enabled). A zero-size hidden `Menu` with `.keyboardShortcut` (`.contextShortcut`) binds the chord `⌃.` (Control-Period) on focused rows/elements across the tree, evoking the `⋯` more-actions menu without conflicting with macOS Cancel (`⌘.`), primary activations (Space/Return), search (`⌘K`), or window commands.

Applied to all 4 `.contextMenu` sites in the tree (`PaletteShelf`, `ColorField+AutomaticMenu`, `SpaceAssignmentChip`, and `SpacesSection`), reusing the exact same menu builder as `.contextMenu` and `.accessibilityActions`.

Draggable elements (`SpaceAssignmentChip`) maintain unobstructed mouse drag behavior via zero-size frame and background placement.

`KeyboardActionParityTests` has been extended to assert that every `.contextMenu` builder in the tree matches both `.accessibilityActions` and `.contextShortcut`. Documented in `docs/user-guide.md` and `.claude/rules/gui.md`.

## Related Issue(s)

Fixes #845

## Checklist

  (see AGENTS.md §3)
  required)
  passes

## How I verified

1. Ran `swift build && swift test && ./scripts/lint.sh` — 3788 tests in 711 suites passed and lint exited 0.
2. Verified that removing `.contextShortcut` from any `.contextMenu` site in the tree causes `KeyboardActionParityTests` to fail with a clear assertion issue (mutation proven red).

## Notes for Reviewer

Uses `⌃.` (Control-Period) as the chord for contextual action menus on focused elements, matching the `.` / `⋯` mnemonic while avoiding `⌘.` (macOS Cancel in dialogs).


hajiboy95:
**Review round complete** (code-reviewer + architect-reviewer +
guard-prover + docs-steward), findings fixed in `8ad93a9e` by
the maintainer. The headline finding was a **blocker**:
`.keyboardShortcut` is window-scoped and SwiftUI resolves
duplicates by hierarchy order, so N identical per-row ⌃.
bindings sent the chord to the FIRST row on the page whatever
held focus — including firing the first color well's
"Automatic" (a reset) from an unrelated control, and the first
tile's Delete cross-targeted. The fix reshapes the seam:

- `rowActions(id:_:)` is now the ONE composition point — the
  builder is handed over once and the seam applies right-click,
  VoiceOver actions and the chord itself, focus-gated through
  `FocusedValues`: each row publishes its identity while its
  subtree holds focus, and only the matching row installs the
  hidden anchor. One live binding, anchored at its row, inert
  with nothing focused. A crossed pairing or mirrored list can
  no longer be expressed at a call site.
- `SpaceAssignmentChip` takes `.focusable()` — it had no Tab
  stop at all, so its claimed keyboard route had no target.
- `KeyboardActionParityTests` reshaped: bare channels banned
  outside the seam (empty exemption map), the seam's
  composition + exact ⌃. chord line + focus gate pinned.
  Guard-prover: 4 mutations red-proofed (dropped seam call,
  bare channel, chord rebind, focus-gate removal).
- Docs made true in the same change set: the design-decisions
  Monitors passage no longer calls the keyboard route
  "knowingly absent"; the #845 ruling has its design-decisions
  entry (preserving the two ⋯-rejection rulings); ui-patterns
  carries the row-menu convention; gui.md's bullet is an
  obligation; the §5 gui row is refreshed.

Gate: 3802 tests / 714 suites green, lint 0, rebased on main.

**Remaining before merge — device pass (owner), keyboard
navigation ON (System Settings ▸ Keyboard), per gui.md this IS
the verification for this feature; nothing headless can see it:**
1. Colours & Motion ▸ My palettes: Tab to a saved tile, ⌃. —
   the menu opens ON THAT tile (not the first), Rename/Export/
   Delete reachable, Escape returns focus to the tile.
2. Two saved palettes: focus the SECOND, ⌃. must open the
   second's menu (the blocker's exact repro).
3. Spaces: Tab into a row's mode picker, ⌃. opens that row's
   menu; `.draggable` still drags.
4. Monitors: Tab reaches an assignment chip at all (new
   `.focusable()`), focus ring visible, ⌃. opens its menu,
   drag still works.
5. A color well with Automatic: focus the hex field, ⌃. opens
   ITS Automatic item; with focus in an unrelated control, ⌃.
   does nothing.
6. Read CI's `Release Build` before merging.

Proposed follow-up (not in this PR, owner to rule): an in-app
hint for the chord — it is currently discoverable only in the
user guide, the ruling's accepted cost. Candidate: the `?`
contextual-help affordance (#94) or the rows' `.help` strings.

hajiboy95:
Re-review of the fix batch converged: **0 blockers, 0 majors, 2
minors**, both fixed and proved in `706c47c3` (the channel ban
now also catches labeled-argument spellings — red-proofed with a
`.contextMenu(menuItems:)` mutation — and the two stale
button-style exemption reasons name the seam). It also
independently confirmed the seam's SwiftUI semantics: descendant
focus publication covers the spaces rows, `SpaceID`
canonicalization rules out id collisions, and the focused-row
anchor cannot oscillate.

One honesty note: a single full-suite run on this branch failed
with 2 issues that did not reproduce in two subsequent clean
runs (3805/714 green, exit 0) — the pipe in my command swallowed
the failing test names, so the flaky pair is unidentified. If CI
reds on this head, the log here is the lead; otherwise treat as
an interleave flake to watch.

Remaining before merge: the 6-step device checklist above
(keyboard navigation ON) and CI on this head.

hajiboy95:
**Device QA round (owner live, 2026-08-23 evening): the chord
now WORKS on both focus worlds** — confirmed on a focused
palette tile (SwiftUI focus) and a spaces row's mode picker
(AppKit focus). It took two mechanism replacements, both
committed as `$(git rev-parse --short HEAD)`:

1. The focus-gated hidden `Menu` never received the key
   equivalent under AppKit-backed focus — the chord beeped.
   Delivery moved to one `NSEvent` local monitor resolving the
   focused row itself (FocusedValues when live, geometric
   containment of the first responder otherwise).
2. A synthesized right-click was nondeterministic (one build
   SELECTED the row instead of opening its menu). The chord now
   presents the same builder's actions in a row-anchored
   popover — deterministic, first action keyboard-focused,
   Escape closes.

Known cosmetic residue (owner: acceptable): on the spaces rows
the popover can anchor visibly away from its row.

Gate: 3805/714 green, lint 0; seam needles re-proved (chord
drift and a dead opener each red the guard). Remaining: CI on
this head, then merge.

Issue #956, opened 2026-08-23
💡 Settings accordions: full-row click and a visible expandable affordance


hajiboy95:
### Use case

Settings accordions (`SettingsDisclosure`) hide options behind a
header that reads as a plain label: the only visual cue is the
small native triangle, and the only HIT TARGET is that triangle
— on macOS, System Settings and Finder disclosure rows are fully
clickable, so this both under-signals ("is this expandable at
all?") and under-delivers when clicked (owner live QA,
2026-08-23). A first-round fix (the `.inline` chrome's top
hairline, owner 2026-08-10, after the App Bar Style accordion
"was nearly overlooked") was not enough — the second round must
change the header itself.

### Proposed solution

Per the `ui-designer` consult (2026-08-23), both asks land at
one seam — a custom `DisclosureGroupStyle` inside
`SettingsDisclosure.swift`:

1. **Full-row click:** the style's label row becomes ONE
   full-width `Button` (`.plain` — an explicit style, which is
   what brings it under the style guards), chevron + label +
   accessory + Spacer, `.contentShape(Rectangle())`, toggling
   `config.isExpanded`. One hit area, one focus stop,
   platform-standard behavior customized the platform's way.
   Preferred over `.onTapGesture` on the label, which leaves two
   hit mechanisms and is invisible to focus and VoiceOver.
2. **Resting affordance:** apply the repo's existing
   ambiguous-control cue — the adaptive hover chip
   (`hoverHighlight()`, 0.06 rest → 0.12 hover) — to the header
   in both chromes, and give the chevron actual weight:
   `chevron.right`, `.footnote.weight(.semibold)`,
   `.secondary`, rotating 90° on expand. Keep the `.inline`
   hairline.

Re-earning what the native triangle gave free
(`LinkedCaptionHitTests` lesson): the Button loses the native
expanded/collapsed announcement, so it owes an
`.accessibilityValue` carrying it back — two new `L()` keys
(`settings.disclosure.expanded` / `.collapsed`), catalogs moved
only via `scripts/*-key(s)`. Space/Return toggle on the focused
header.

Also: `SpaceOverrideRows+Footer.swift:82` holds the one bare
`DisclosureGroup` left in the tree — it takes the same style in
the same change or becomes the odd header out.

### Alternatives considered

- `.contentShape` + `.onTapGesture` on the label: less code, but
  two hit mechanisms and a gesture path invisible to
  focus/VoiceOver.
[...]

Issue #760, opened 2026-08-06
💡 Advanced reveal: show which cards changed and which are new


hajiboy95:
### Use case

Switching from Simple to Power User changes the page under the
user without saying what it changed. Rows appear, cards appear,
some cards grow — and the user has to re-read the whole page to
find out what the toggle did. The first time, most people flip it
back and forth to compare, which is the tell that the transition
carries no information.

The mode is the app's main depth affordance (`SettingsArea.minimumMode`,
`SettingTier.showMore`), so the moment it fires is worth making
legible.

### Proposed solution

**Animate what changed on the way in.** When the mode flips to
Power User, briefly highlight the rows and cards that are new to
the page — a short glow that decays, and nothing after it. Enough
to draw the eye to three places at once; gone before it becomes
chrome.

**Give advanced content a persistent, quiet tier signal** so the
answer survives the animation: the `.showMore` rows already have
a tier, and it can carry a slightly stronger border than the
immediate rows.

**Not a second accent hue.** The proposal to paint advanced
content in a different green should not ship:

- The theme has one accent. A second saturated hue in the same
  family reads as a *state* (ok / warn), not as a tier, and this
  window already spends green on control fills.
- Colour alone cannot carry the distinction under the
  colour-vision clause, so the border weight has to do the work
  regardless — at which point the second hue only adds a way to
  be misread.
- `SettingsThemeTokenTests` holds the one copy of the hex table
  and resolves every token in light and dark; a second accent is
  a token pair plus a wiring lens for a signal the border weight
  already carries.

So: motion answers "what just happened", border weight answers
"what is advanced", and hue answers neither.

### Alternatives considered

- **A count in the header ("6 more settings").** Cheap and
  honest, but it does not say *where*, which is the actual
  question.
- **Animate the newly-hidden rows on the way back to Simple
  too.** Probably not: on the way out the rows are leaving, and
  drawing attention to content that is about to disappear is
  noise. A plain fade is enough there.
- **Leave the transition instant.** The status quo; it is what
  makes people toggle twice.

### Extra context

Motion accessibility: honour Reduce Motion — with it on, the
reveal is instant and the border-weight tier signal alone carries
the answer. That is the argument for the persistent signal
existing at all rather than relying on the animation.

Targeted at #678 Phase 4 (Easy/Advanced mode mechanics).

hajiboy95:
[...]

PR #988, opened 2026-08-25
fix(gui): open a drawer without animating the opening

### LINKED ISSUE #961: 🐛 Opening a Settings disclosure makes VoiceOver repeat its previous announcement

0.9.9 dev, branch `feat/voiceover-hardening` (#812 device sessions).

### macOS version

macOS 26.6.2

### Mac model type

### Steps to reproduce

1. VoiceOver on, open Settings and go to an area with disclosure groups (e.g. Colours & Motion).
2. Let VoiceOver announce something (a row, a value).
3. Open an accordion (a `SettingsDisclosure`) by click or keyboard.

Expected: opening the disclosure announces the disclosure itself (expanded/collapsed), or nothing beyond it.

Actual: VoiceOver re-announces whatever it had said previously — the last announcement repeats. Focus is not stolen (verified during the #812 sessions); only the speech repeats. Owner report 2026-08-24: a nit, not disruptive.

### Relevant output and logs

Not applicable — no runtime failure. Suspected mechanism: the disclosure's expansion rebuilds part of the subtree and VoiceOver re-reads the element under its cursor after the layout change. `SettingsDisclosure` (Sources/KiwiDesk/Settings/Components/Common/SettingsDisclosure.swift) is the component; whether the repeat comes from the AX tree invalidation or from an `AccessibilityNotification` echo needs an Accessibility Inspector session.

- [x] None that I know of

### Extra context

Found during the #812 VoiceOver hardening device sessions and deliberately not folded into that lane (owner ruling 2026-08-24: "nothing important for now", explicitly NOT on the 1.0 milestone). Filed so the observation is not lost.

hajiboy95:
## Description

Opening a Settings disclosure made VoiceOver **re-announce
whatever it had last said** — a row, a value, anywhere on the
page. Focus was never stolen; only the speech repeated.

That symptom is the whole diagnosis. A disclosure announcing
itself would say "expanded"; what was heard instead was the
*previous* element, which is the signature of VoiceOver
**re-resolving the element under its own cursor** after the
accessibility tree moves beneath it.

**Mechanism.** Inserting `configuration.content` rebuilds the AX
element list under the hosting view. That part is unavoidable —
it is what opening a drawer *is*, and no arrangement of the
style avoids it. What made it expensive was running the toggle
inside a transaction: the rebuild then stretched across every
frame of the 0.18 s easing, so every element below the drawer
inside the detail `ScrollView` had a **moving accessibility
frame for the whole animation**. VoiceOver answers a moving tree
by speaking its cursor again.

**Fix.** The motion is not the price of fixing that; only the
animated *layout* is. `rotationEffect` takes no layout and the
chevron is already `.accessibilityHidden(true)`, so the same
easing scoped to the chevron alone — `.animation(_:value:)`,
not a transaction — leaves the tree changing exactly once while
keeping the cue #956 ruled. The header stays a full-row `.plain`
`Button` announcing expanded/collapsed via `.accessibilityValue`;
nothing in that ruling moves.

**The cost, stated plainly: the drawer's contents now appear
rather than arrive.** That is the trade — a fade for a sighted
reader against a repeated announcement on every drawer, every
time, for a VoiceOver reader. `docs/design-decisions.md` carries
it as a **question**, not a ban — *what is the reflow buying?* —
because the unqualified form ("animate what does not move the
tree") contradicts the standing #760 ruling that the mode flip's
pane *does* animate its insertion, which is guard-pinned by
`ModeGatedChromeTests`. That ruling stands. What this adds is
the price tag it was made without, and the residue that the mode
flip is animating an insertion into the same pane and so very
likely carries the same repeat.

The obligation also goes into `.claude/rules/gui.md` and the
AGENTS.md §5 row rather than only onto the docs shelf: the
author about to animate an insertion has the rule file loaded
and `docs/design-decisions.md` nowhere. And it goes onto
`externalExpansion`'s doc comment, because 14 call sites write
that binding and a `withAnimation` around one of them
re-introduces the defect from outside the file the guard reads.

Candidates rejected, with reasons, rather than left unsaid:
`.id()` on the content (identity cannot stop an absent view from
being inserted); `.accessibilityElement(children: .contain)` on
the style's `VStack` (the change is already inside that stack,
and it inserts a navigation group the ruling did not ask for);
keeping the content mounted and toggling `.accessibilityHidden`
(changes the element list by exactly as much, while paying full
layout for collapsed content and risking collapsed controls in
the tab order).

## Related Issue(s)

Closes #961

## Checklist

  the app to confirm it works as intended — **this is the open
  item; see "How I verified"**
  (see AGENTS.md §3)
  required)
  passes
- [x] Release build green — skipped locally and left to CI's
  `Release Build` job: no concurrency or `Sendable` change

## How I verified

- `swift build`, `scripts/lint.sh` exit 0, full suite green
  (**3878 tests in 730 suites**).
- `SettingsDisclosureMotionTests`, a new suite, pins both halves
  and was proven red by `guard-prover` against a full revert and
  each half-revert separately, plus a vacuity probe (a source
  scan that matches nothing must fail, not pass — one clause did
  pass on empty input, which is why three positive nets sit
  beside it).
- After review widened it, re-proven: full revert → 2 failures;
  a `.transaction { $0.animation = … }` on the button → caught
  (it was not, before); retuning the easing's duration → still
  green, so the guard carries no false red on a taste change.

**What is NOT verified, and cannot be from here: that VoiceOver
actually stopped repeating.** The new suite is a source scan —
it says the transaction is gone, never that the speech changed.
`SettingsDisclosureHeaderTests`' own docstring already states
that limit for this component, and it binds here.

**So this PR wants a device session before it merges**, and the
session has a second job beyond confirming the fix:

1. VoiceOver on, Settings → Colours & Motion. Let VoiceOver
   announce a row, then open a drawer. The repeat should be
   gone.
2. **Repeat it with Reduce Motion ON.** This is the
   discriminator, not a formality: Reduce Motion already routed
   the old code to `withAnimation(nil)`, so if the repeat used
   to happen with Reduce Motion on as well, the animation was
   never the whole mechanism and this fix is at best partial.
3. Try a `.card` drawer (an "Advanced" one) as well as an
   inline one — they take different chrome and only the inline
   form publishes a scroll anchor.
4. Eye-confirm the trade: the contents now pop rather than
   fade. If that reads as broken rather than brisk, say so —
   the ruling is worth revisiting, the mechanism is not.

## Notes for Reviewer

The diagnosis rules out three other candidates on evidence, not
taste, and they are worth knowing if the device session says the
repeat survives:

- **The `.accessibilityValue` flip** posting a value-changed
  notification — rejected: macOS speaks a value change only for
  the element under the cursor, and the cursor is elsewhere in
  this repro. If this were the driver you would hear "expanded"
  injected, which is not what was reported.
- **The `HoistedRevealAnchorsKey` preference** forcing a parent
  rebuild — rejected: its value is `Equatable` and unchanged by
  expansion, and the overlay it drives is zero-size `Color.clear`
  that vends no AX element.
- **Hover state** — rejected: it reproduces by keyboard, where
  `HoverChip`'s state never changes.

One rider deliberately NOT taken: `.accessibilityValue` sits
after `.rowHoverHighlight` in the chain. That is safe today
because `HoverChip` is only `.padding` + `.background`, neither
of which forms an element — so the value lands on the Button.
It is load-bearing by accident rather than by construction, but
moving it is a separate change and not this fix.


hajiboy95:
Closing unmerged: the hypothesis this PR is built on is refuted by a device session (owner, 2026-08-25).

Removing the expansion's transaction does NOT stop the repeat. And the symptom is not what #961 describes: VoiceOver's cursor does not merely re-announce in place — the pane **scrolls back** to the focused element and reads it there, while keyboard focus itself never moves.

The decisive measurement: **with VoiceOver off there is no repro at all.** The detail pane's scroll driver (`SettingsView+Detail.reveal`) runs identically with and without VoiceOver, so it is exonerated; what remains is VoiceOver re-resolving its own cursor after the accordion changes the accessibility tree, scrolling that element into view and announcing it.

That is a property of the tree changing, which is what opening a drawer IS, so no arrangement of `SettingsDisclosureStyle` fixes it — exactly the branch the diagnosis flagged as 'the fix is not in these two files'.

Not merging a change that costs the drawer's open/close animation and buys nothing measurable. Findings recorded on #961; the branch is deleted, the evidence is the durable part.

Issue #1013, opened 2026-08-25
💡 Scheduled update alerts go unnoticed: no gentle reminders for a menu-bar app


hajiboy95:
### Use case

I want to be told when a KiwiDesk update is available without
having to go looking for it. KiwiDesk checks in the background on
purpose — `docs/design-decisions.md` ▸ *Background update checks
are on, and there is no switch* argues that an updater nobody
remembers to run is not an update path — so the background check
is the path most users are actually on.

But KiwiDesk runs permanently in **accessory** mode
(`docs/design-decisions.md` ▸ *Permanent accessory mode*), and
that is exactly the shape Sparkle warns about. For a background
app, `SPUStandardUserDriver` deliberately shows a **scheduled**
update alert *without* activating — behind other running apps
when the app is inactive, and `orderBack:` even when it is
active and has a key window (`SPUStandardUserDriver.m`
`setUpActiveUpdateAlertForScheduledUpdate:`, Sparkle 2.9.6). It
only activates when the check lands within ~3 s of updater
initialization.

So the alert is drawn and, for a menu-bar app with no Dock icon
to bounce, effectively invisible. Sparkle knows this and logs it
once per launch:

> Warning: Background app automatically schedules for update
> checks but does not implement gentle reminders. As a result,
> users may not take notice to update alerts that show up in the
> background.

KiwiDesk builds its user driver with no gentle-reminder support,
so it is squarely in that warning.

### Proposed solution

Implement Sparkle's gentle reminders
(https://sparkle-project.org/documentation/gentle-reminders) on
the existing `UpdatePromptPolicy`
(`Sources/KiwiDesk/Updates/AppUpdater.swift`), which is already
the `SPUStandardUserDriverDelegate` and already the one place
this app answers Sparkle's UI questions:

- `supportsGentleScheduledUpdateReminders` → `true`
- `standardUserDriverShouldHandleShowingScheduledUpdate(_:andInImmediateFocus:)`
  → `false` when not in immediate focus, so KiwiDesk shows the
  reminder itself
- `standardUserDriverWillHandleShowingUpdate(_:forUpdate:state:)`
  → post the reminder
- `standardUserDriverDidReceiveUserAttentionForUpdate(_:)` and
  `standardUserDriverWillFinishUpdateSession()` → withdraw it

The open design question is **which channel the reminder uses**,
and it should be ruled before any code:

1. **A mark on the menu-bar item** — a badge or a changed glyph
[...]

PR #1014, opened 2026-08-25
fix(updates): bring the install-and-restart prompt forward

### LINKED ISSUE #1011: 🐛 The install-and-restart window hides behind other windows

`1.0.1` — reported against the shipped 1.0.1 build during live
QA on 2026-08-25.

### macOS version

macOS 26.6.2 (25G83).

### Mac model type

### Steps to reproduce

1. With an update available, open the menu-bar menu and choose
   **Check for Updates…** (Sparkle's update window appears in
   front — `checkForUpdates` calls
   `NSApp.activate(ignoringOtherApps:)` first).
2. Click **Install Update** / **Install and Relaunch**.
3. Look for the install-and-restart window.

**Expected:** the window that asks to install and restart stays
in front, the way the update window that preceded it did — it is
the one thing the user has to answer.

**Actual:** it drops **behind every other window**. Nothing tells
the user it is waiting; the update looks stalled until they find
the window by hand.

Not fatal — the update still applies on the next restart — but it
makes the in-app update path read as broken, which is the path
`docs/design-decisions.md` ▸ *No distribution channel without an
update path* exists to keep trustworthy.

### Relevant output and logs

Not captured — the defect is window ordering, and the reporting
run was a manual update on the shipped build.

The likely mechanism, for whoever picks this up: KiwiDesk runs
permanently in **accessory** mode (`docs/design-decisions.md` ▸
*Permanent accessory mode*), so its windows do not get the
frontmost treatment a regular app's do. The one activation is at
CHECK time (`StatusItemController+Updates.swift`,
`checkForUpdates`), which covers the window Sparkle opens then —
but the install/relaunch prompt comes later, and Sparkle's own
installer UI runs out of `Sparkle.framework`'s bundled
`Updater.app` / XPC services rather than out of this process, so
that activation never reaches it. Worth checking whether
`SPUStandardUserDriverDelegate`'s
`standardUserDriverWillShowModalAlert` (or
`…ShouldHandleShowingScheduledUpdate`) is the seam that re-raises
it, and whether the accessory policy needs a scoped exception for
that one window.

- [x] None that I know of

### Extra context

- Reported by the owner during the #1010 QA session, on a
  two-screen setup ("Displays have separate Spaces" ON). Whether
  the second screen is involved is unknown — worth reproducing on
  one screen first, since the accessory-mode explanation above
  needs no second display.
- Related: the packaging half of the update path is
  `.claude/rules/packaging-and-release.md`; the Sparkle wiring
  seam is guarded by `UpdaterSeamGuardTests`, which pins the
  construction and wiring but says nothing about window ordering.

hajiboy95:
## Description

Sparkle's install-and-restart prompt appeared behind every other
window, so the in-app update path read as stalled.

**The mechanism was not the one the issue guessed.** Sparkle does
activate for the windows it opens on a check the *user* asked
for. The install prompt is not another window: it is a later
state of the status window the **download** opened, and Sparkle
announces that state with `NSApp.requestUserAttention` alone —
which bounces a Dock icon a permanent-accessory app does not
have. So the ordinary path (start the download, go back to work
while it runs) ended with the prompt behind everything, with
nothing saying the update was waiting.

Neither seam the issue proposed reaches that moment:
`standardUserDriverWillShowModalAlert` fires only for modal
`NSAlert`s, and the nearest updater-delegate hook
(`didExtractUpdate`) fires while the installer is still
preparing. `UpdatePromptDriver` overrides the user driver's
`showReadyToInstallAndRelaunch`, which names the moment instead
of approximating it — and because `SPUStandardUpdaterController`
builds its stock driver inside its own initializer with no
injection point, `SparkleUpdater` now assembles `SPUUpdater` plus
that driver itself.

**Activation alone was not enough**, which review caught. It does
not deminiaturize, and Sparkle offers that status window a
minimize button unless its delegate refuses — the window the user
parks during the download *is* the one that becomes the prompt,
and there is no Dock icon to recover it from.
`UpdatePromptPolicy` refuses the affordance, and carries the
modal-alert activation too, so Sparkle's error and
acknowledgement alerts come forward as well.

## Related Issue(s)

Closes #1011. Filed from this lane: #1013 (scheduled update
alerts go unnoticed — see *Notes for Reviewer*).

## Checklist

  (see AGENTS.md §3)
  required)
  passes

## How I verified

**Device round DONE 2026-08-25, by the owner, on macOS 26.6.2.**
Both routes below confirmed against a real signed 1.0.0 → 1.0.1
update off the live appcast. Evidence is in the first comment.

What is done: `swift build`, `swift build -c release`,
`swift test` (3932 tests in 740 suites) and `scripts/lint.sh` all
green, re-run at every commit.

What is prepared for the device round — two `.app`s, both stamped
**1.0.0** so the live appcast offers them the published 1.0.1,
both signed with the release Developer ID so the Accessibility
grant carries over:

- `broken/` — `main` as it stands (34bb8a39)
- `fixed/` — this branch

Two routes, and both matter:

1. *Check for Updates…* → **Install Update** → ⌘-Tab away
   immediately → wait ~10 s without clicking. On `broken` the
   prompt is behind the app you switched to; on `fixed` KiwiDesk
   comes forward with it.
2. Same, but **minimize** the download progress window instead of
   switching away. On `broken` the prompt is unrecoverable; on
   `fixed` the minimize button is not there.

Route 2 is the one that separates the final fix from its first
version — an activation-only fix passes route 1 and fails route
2. The ⌘-Tab in route 1 is not optional either: the download plus
install preparation is only a few seconds, and if KiwiDesk never
loses focus both builds look identical.

## Notes for Reviewer

**Three things review changed, worth knowing before reading the
diff.**

*The scoping.* The design-decisions corollary first stated a flat
obligation — "a window the user must answer activates at the
moment it appears" — which the tree does not meet on the
**scheduled** update path, where Sparkle deliberately shows the
alert behind other windows for a background app. Widening the
rule to cover that would contradict *Background update checks are
on, and there is no switch* one section up. It is now scoped to
an interaction **already begun**, with the opposite rule stated
for an unsolicited offer. That cut is Sparkle's own, not one
drawn to fit the code: a scheduled check passes
`showErrorToUser:_showedUpdate` where a user-initiated one passes
`YES`. The residue — a background app's scheduled alert being
easy to miss, which Sparkle answers with *gentle reminders*
KiwiDesk does not implement — is **#1013**, not ruled acceptable
here.

*What we now own.* Dropping `SPUStandardUpdaterController` means
owning the updater start and its failure path. A refused start is
logged and shows nothing, where the controller put up a modal
`NSAlert` — developer-facing by Sparkle's own description, and an
unactivated alert from a process with no Dock tile is the same
defect this PR fixes. The user-facing signal is the greyed *Check
for Updates…* row, since `canCheckForUpdates` only turns true
inside a successful start. Config Issues is **not** available as a
channel: that model's list is replaced wholesale by
`KiwiCore.onConfigIssuesChange`, so a GUI-side entry would be
wiped. The ruling is stated in the init doc rather than left
implicit.

*Guards.* Four `guard-prover` rounds, ~40 mutations, and they
found three things no review did: a rewired `userDriver:` that
left the override as dead code with every count still at one; a
decoy class that satisfied an unscoped body scan; and a needle
whose word `false` was satisfied by a body returning `true`. The
last is why the policy's answer is asserted by **calling** it
rather than scanning for it — and both its delegate methods are
*optional* ObjC requirements, so a rename compiles and silently
restores the minimize button, which the `responds(to:)` half
catches.

Two limits are stated in the suites rather than left to be
discovered: a source needle reads a spelling, so a call behind a
never-true condition passes; and the wiring needles hold the
spelling, not the object identity — a `typealias` plus a
shadowing local still passes. Closing the second needs a
type-level check (an `SPUUpdater` built against a fake driver),
which is a design change rather than a needle.

**One residue named in the code**, since it is real and narrow:
refusing minimization closes the *parking* route only. `super`
reuses a status controller it already has rather than re-ordering
it, and activation preserves intra-app window order — so a
Settings window opened during the download still sits above the
prompt. Rarer than the reported shape, and Sparkle hands the
status window to nobody, so closing it would need a handle we do
not have.

https://claude.ai/code/session_016xfTeAoeY6iM6dcLEfD2U3


hajiboy95:
## Device round — 2026-08-25, owner's machine, macOS 26.6.2

*(This comment was rewritten twice as the owner's testing
corrected me. What follows is the final state; the earlier
versions claimed a mechanism and a result that were both wrong.)*

Run against a real signed update off the live appcast: two
`.app` builds stamped **1.0.0** (so the feed offers them the
published 1.0.1), both signed with the release Developer ID.

### The reported mechanism was not the one I first fixed

I diagnosed the moment the prompt *appears* and fixed that. The
owner's report — "even when I don't click anywhere, focus
immediately went to the window behind" — is one step earlier, and
it is the visible half:

`SUUpdateAlert.m:128` closes the alert window **before** invoking
the completion block on line 131 that starts the download.
Between those two lines an accessory process has no windows at
all, so macOS deactivates it and hands focus to whatever was
behind. Sparkle then creates the status window into an app that
is already inactive, and it opens behind. **Clicking *Install
Update* is enough on its own; no switching away is involved.**

My first QA instructions insisted on ⌘-Tab-ing away quickly,
which is also why an earlier run here was recorded as "did not
reproduce". It had reproduced; the instructions were pointing at
the wrong thing.

### Confirmed, `broken/` (main at 34bb8a39) vs `fixed/`

| Route | `broken/` | `fixed/` |
|---|---|---|
| Click **Install Update**, then do nothing at all | window disappears, focus jumps to the window behind | progress window stays in front |
| Click, then ⌘-Tab away and wait | prompt never appears | KiwiDesk comes forward with it when ready, and not before |
| Click, then minimize the progress window | prompt unrecoverable | window has no minimize button at all |

Route 1 is the owner's actual report and the one that needed the
second override. Routes 2 and 3 still matter and are not made
redundant by it: going elsewhere during the download is ordinary,
and parking the window is something no activation can undo.

Screenshot of the prompt on the owner's machine shows it with no
traffic lights at all — Sparkle sets the ready-to-install state
non-closable itself, and `UpdatePromptPolicy` refuses the
minimize half, so the window can only be answered.

### Install-on-quit, confirmed

Quitting a copy with an update staged applied it without the
prompt ever being answered — `broken/` came back as
`1.0.1 (555d8a9)`. That is Sparkle's design, untouched here, and
it is what #1011 meant by "not fatal". It defines the defect: the
update was never at risk. What was broken is that nothing told
the user it was waiting, and the app did not relaunch at the
moment they chose. The user guide now says this outright.

### Owner verdict

> yeah works now perfect […] it's much better than before

### Observed and NOT fixed here

**Tiled windows shift slightly when the activation lands.** The
owner's words: "windows move a bit now in the background but not
severe." This is a side effect of *this* change — activating
KiwiDesk makes it frontmost, and something in the focus/raise
path reacts. Judged not severe by the owner and not blocking the
merge, but it is a real observation caused by this PR rather than
a pre-existing one, so it is filed rather than left in a thread:
see the follow-up issue linked below. It needs a log capture to
characterize, which this round did not take.

hajiboy95:
Follow-up filed for the window shift observed in the device round: #1015 (Bug, Low/Medium, no milestone). It is caused by this PR's activation, so it is recorded rather than left in a thread — not blocking, per the owner's judgement that it is not severe.

Issue #1023, opened 2026-08-25
🐛 The Desktop switch moves the space pointer without performing the transition


hajiboy95:
### KiwiDesk version or commit

`9b8b3904` (`main`, carrying #1008's Desktop verbs and #1010),
built as a signed `.app`. Also reproduced on the `#1007` branch —
that branch is not the cause; see *Extra context*.

### macOS version

macOS 26.6.2 (25G83).

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Two screens, "Displays have separate Spaces" ON. (Also
   reproduces with the target Desktop on the SAME screen.)
2. Focus a window.
3. `kiwidesk move_to_desktop_and_follow <N>` where N is a Desktop
   that is **not currently shown**.

### Expected vs. actual behavior

**Expected:** the screen holding Desktop N switches to it, the
window is there, and KiwiDesk's state matches what is on screen.

**Actual:** **the switch does not happen, and KiwiDesk believes
it did.**

- macOS keeps showing the previous Desktop.
- The moved window's membership DOES change — Mission Control
  shows it on Desktop N — but it keeps being **rendered over the
  Desktop still on screen**, as an overlay.
- KiwiDesk proceeds as though the switch landed: it tiles for the
  destination space, the Space Bar updates, and the layout is the
  new Desktop's.
- Swiping to Desktop N by hand makes everything correct at once —
  the window tiles properly and is no longer over the old
  Desktop.

Owner's words: *"space isn't really changed but it behaves like
it's in the new space until we really go there"*.

### Relevant output and logs

Captured 2026-08-25, `subsystem == "com.kiwicanopy.kiwidesk"`,
five consecutive invocations:

```
22:55:42.860  follow: owing focus to w131136 when its Desktop reveals it
22:55:46.077  follow: owing focus to w131136 when its Desktop reveals it
22:55:50.588  follow: owing focus to w131136 when its Desktop reveals it
22:55:52.650  follow: owing focus to w131136 when its Desktop reveals it
22:55:55.086  follow: owing focus to w131136 when its Desktop reveals it
```

**What is absent is the evidence.** Across all five: no window
destroy, no `close-return` decision, no Desktop change, no
reconcile — and, decisively, **no `the Desktop bridge refused the
switch` line**. `WMBridge.setCurrentSpace` returned **success**
and nothing happened.

### The mechanism, and the rule it violates

`KiwiCore.switchDesktop` (now
[...]

PR #1036, opened 2026-08-26
feat(commands): a follow carries focus to the window it moved

### LINKED ISSUE #1007: 🐛 A follow does not carry focus onto a hidden Desktop
### Use case

`move_to_desktop_and_follow(n)` promises to send the focused
window to a Desktop **and take you with it**. With more than one
screen it keeps only half of that: the window moves, the
Desktop's own screen switches to show it, and your keyboard
focus stays on the screen you were already using. You watch the
window arrive somewhere else.

The cause is not a bug in the verb — macOS attaches keyboard
focus to a *window*, never to a screen, so "switch that screen's
Desktop" is the whole of what the switch can do. Carrying the
user across means focusing the moved window once its Desktop
reveals it, and at the moment of the move that window is not
addressable: it is on a Desktop nobody is showing, so
Accessibility does not list it.

Single-screen setups never meet this: the Desktop that switches
is the one you are on, and the reveal's own focus handling lands
you correctly (device-checked 2026-08-25 during #884's lane).

### Proposed solution

Give the follow path a **pending focus**, drained where the
window becomes addressable — the same shape #890 already plans
for a composed `space:` target, and the natural companion to it:

1. `move_to_desktop_and_follow` records "focus window W when
   Desktop D reveals" beside the move.
2. The reveal reconcile — `desktopSettle`, which already
   re-asserts focus 600 ms after a switch — drains it: if W is
   now tracked and listed, focus it instead of the Desktop's
   remembered anchor.
3. The record expires if that Desktop is not shown within a
   bounded window, so a follow that macOS declined cannot fire
   minutes later.

Open sub-question for whoever picks this up: whether focusing a
window on another screen is the whole answer, or whether the
mouse should warp to it as `focus_space` does for its own
switches (`warpMouseToFocused`). That is a product call, not a
mechanism one.

### Alternatives considered

- **Do nothing and document it.** What shipped in #884's lane —
  the Lua reference and the design-decisions entry both say the
  follow does not carry focus across screens. Honest, and enough
  while the verbs are new, but it leaves the verb's own name
  over-promising.
- **Refuse a cross-screen follow.** Worse: the move is the part
  the user asked for, and refusing it to protect the follow's
  wording would lose the useful half.
- **Activate the app instead of focusing the window.**
  `activate()` picks whatever window that app last had, which on
  a multi-Desktop setup is regularly the wrong one — the same
  reason `focus_space` raises a specific window rather than
  activating an app.

### Extra context

- Shipped in the #884 lane; the mechanics are recorded in
  `docs/lua-reference.md` ▸ *move_to_desktop_and_follow* and in
  `docs/design-decisions.md` ▸ *The window-management bridge is
  not a SIP escape hatch*.
- **Not device-verifiable on the maintainer's machine today** —
  it has one screen. Anyone picking this up needs two, plus
  "Displays have separate Spaces" on.
- Related: #890 (the collector; its pending-assignment item is
  the same machinery), #884 (the verbs), #888 (why a Desktop
  number is global rather than per screen).

hajiboy95:
## What

Fixes #1007 — `move_to_desktop_and_follow` now carries keyboard
focus to the window it moved, on every arm of the switch:

- **Hidden target** (`.switched`): the follow records a focus
  debt (`FollowFocusIntent` — per-window, rekey-following,
  bounded at 5 s so a follow macOS declined cannot fire late)
  and the window's own ARRIVAL pays it — the moment it becomes
  addressable again — as a space switch (activate + focus +
  emit), never a bare focus. Both ends narrate (`owing focus` /
  `focus handed`), so a trace with the first and not the second
  is a window that never came back.
- **Already-shown target** (`.alreadyShown`): the OS half was
  always right (the window stays visible and keeps key focus);
  the STATE half was not — the re-home stamps the destination's
  focused member but never activates it, so the ring and every
  `focusedWindowID` consumer stayed on the origin space until a
  click (owner-reproduced on device). Both arms now route
  through one `handFollowFocus`.
- **Refused switch** (`.refused`): the move already happened, so
  the follow degrades to the no-follow bookkeeping — latch,
  vanish stamp, reap (review round 1).

Ported from the parked pre-#1023 branch
`fix/cross-screen-follow-focus`, reconciled rather than rebased:
its own close-return stand-down arm and `followedAway` seam are
deliberately NOT ported — the departure is now #1023's
synchronous eager fold, which the `eagerDepartureInFlight` latch
already stands down, and the registration release means no
organic removal reaches the fold afterwards. `switchDesktop`
gains `DesktopSwitchOutcome`; the move verbs split into
`KiwiCore+DesktopMove.swift` at the §2.1 ceiling.

## Verification

- Device: the hidden-target carry is owner-verified ("only works
  when native space actually changes" — i.e. the debt path
  works); the already-shown fix was built FROM the owner's
  device report and awaits their re-test; the refused arm is
  unit-tested only (no way to make the bridge refuse on
  device).
- `guard-prover`: 16 mutations round 1 — 16 red-proofed, 0
  inert, including the parked branch's historical escape (the
  unconditional record) now redding on exactly its designed
  scan; round 2 on the review-fix batch in flight.
- Review: code-reviewer + architect-reviewer + docs-steward
  round 1 — all findings addressed (both architect majors are
  the `.alreadyShown`/`.refused` rulings above; docs-steward's
  three prose amendments ride along).
- Full gate on the rebased tree: 3985 tests, lint, release
  build (via `build-app.sh`) — green.

## Docs

`docs/lua-reference.md` and `docs/cli.md` now promise the carry;
`docs/design-decisions.md` argues the follow/plain-switch
asymmetry and the pointer non-decision (`mouse.follows_focus`
governs, no exception); `state-and-layout.md` + AGENTS.md §5
carry the rule row (`FollowFocusSeamTests` is the wiring
register, `DesktopFollowTests` the end-to-end path).

fixes #1007

https://claude.ai/code/session_014y25qGTmuBr3dYKoR63W6G


hajiboy95:
## guard-prover — final report (both rounds)

**Round 1** (16 mutations): 16 red-proofed, 0 inert, 0 vacuous —
including the parked branch's historical escape (the
unconditional record), now caught by
`FollowFocusSeamTests/onlyASwitchRecords`, and the rekey-wiring
deletion producing exactly one issue in a 3985-test full run
(the register is the sole net, as designed).

**Round 2** (7 mutations over the review-fix batch): 6
red-proofed, 0 inert, 1 accepted residue confirmed:

- The recorder relocated into the `.alreadyShown` arm reds three
  independent nets (in-arm, nowhere-else, and behaviorally).
- The `.alreadyShown` arm's `handFollowFocus` deletion, the
  hand-off's `activate` removal, and the payer inlining an
  identical trio each red exactly one designed clause with every
  behavior suite green — the seam scan is the sole net for the
  space-switch shape, at three pinned sites.
- The refused arm's `departWithoutFollowing` deletion reds both
  new latch/reap asserts.
- The re-home wiring scan still bites after its repoint at the
  split file, and fail-shuts on a stranded path.
- **Accepted residue, by design:** the `.alreadyShown` arm's
  same-screen skip (`destination != activeSpace`) is enforced by
  nothing — dropping it would emit one spurious `space_change`
  on a same-screen follow. A scan clause on the expression would
  be the value-pin tax tests.md bans; the argument lives in the
  arm's own comment, and review owns it.

Tree restored after every mutation; final runs green.

Issue #1075, opened 2026-08-28
💡 Move resize to a dedicated ⌥⌘ size layer: ⌘ means two things, and the one held verb is on the heaviest chord


hajiboy95:
### Use case

Two problems with the current shortcut map, neither of which more
keys would fix.

**`⌘` means two unrelated things.** The #270 ladder is `⌃⌥` →
`⌃⌥⇧` → `⌃⌥⌘`, and tier 3 holds one positional verb and one size
verb:

- `⌃⌥⌘` + arrow → resize
- `⌃⌥⌘` + digit → move to space & follow

So `⌘` reads as "resize" when you press an arrow and "and follow"
when you press a digit. Tiers 1 and 2 are clean — `⌃⌥` acts on
*me*, `⌃⌥⇧` acts on *the window* — and tier 3 is a bag. That is
not only cosmetic: focus, swap, move and follow are all position
verbs over the flat array, while resize changes a weight or a
scroll-slot domain. Resize is the one verb class that does not
belong on a positional ladder at all.

**The one held verb sits on the heaviest chord.** #1056 made
resize the only verb you hold down, and it lives on a three-
modifier chord with `⌃` and `⌘` at opposite ends of the left
cluster. `⌥⌘` is a single thumb roll — the two are adjacent, so
one thumb covers both and leaves the fingers free for the key.
The held verb should be on the lightest chord; today it is on the
heaviest.

**The arrows are ambiguous for resize anyway.** An arrow carries
two competing readings — "which axis and sign" (today's model)
and "which way the edge moves" (what the screen actually shows).
In a tiling WM the free edge depends on where the window sits in
the array, so `⌃⌥⌘←` grows a right-column window and shrinks a
left-column one. Relabelling cannot fix that; the arrow shape
creates it.

### Proposed solution

Move resize off `⌃⌥⌘` onto a dedicated `⌥⌘` **size** layer, keyed
by digits rather than arrows.

| Layer | Meaning | Arrows | Digits |
|---|---|---|---|
| `⌃⌥` | act on **me** | focus | go to space |
| `⌃⌥⇧` | act on **the window** | swap | move to space |
| `⌃⌥⌘` | **both** | *(free)* | move & follow |
| `⌥⌘` | **size** | — | `4`/`5` width, `7`/`8` height |

This is strictly simpler than today, not merely different: the
`⌃⌥` ladder becomes purely positional with a consistent
[...]

PR #1077, opened 2026-08-28
feat(keys): make the numeric keypad bindable (#1074)

### LINKED ISSUE #1074: 💡 No keypad key can be bound to anything
### Use case

I use a full-size keyboard with a numeric keypad. I want to bind
KiwiDesk verbs to keypad keys — most obviously to reach my spaces
with `⌃⌥` + keypad digits the way I already do with the number
row, and to use the keypad's 2×2 block (`7`/`8` sitting directly
above `4`/`5`) for window resizing, where the geometry makes an
axis mapping self-evident that a flat row cannot.

Today none of that is possible: **no keypad key can be bound to
anything at all.** A keypad press matches no binding and does
nothing, for every verb.

### Proposed solution

Make the keypad bindable, and treat its digits as the same keys
as their number-row twins.

**1. Add the keypad key codes.** `KeyCombo.keyCodes`
(`Sources/KiwiDeskCore/Keys/KeyCombo.swift`) stops at the number
row — `"4": 21, "5": 23, "7": 26, "8": 28`. The keypad is a
separate set of Carbon virtual key codes (`kVK_ANSI_Keypad0`…`9`
= 82, 83, 84, 85, 86, 87, 88, 89, 91, 92, plus `+` 69, `−` 78,
`*` 67, `/` 75, `.` 65, `=` 81, enter 76, clear 71) and none of
them are in the map.

**2. Alias the ten keypad digits to their number-row twins**, so
a binding authored as `4` fires from either physical key. The
rule, stated as a closed set:

> The ten keypad digits are the same key as their number-row
> twin. Every other keypad key (`+ − × ÷ . = enter clear`) is its
> own bindable key.

Keep that closure. Do **not** generalise it to "the keypad
mirrors its character" — keypad `+` has no number-row twin
(main-block `+` is `⇧=`), so the uniform version needs an
exception immediately and the digit-only version never does.

Why alias rather than add ten distinct keys: character-based
matching is already the Mac norm — AppKit menu key-equivalents
match on the character, so `⌘1` and `⌘`+keypad-`1` are one
shortcut in essentially every Mac app. KiwiDesk differs only
because it binds by key code, and that choice exists for
international position-stability, not to tell a keypad from a
number row. Aliasing also means every existing user's `⌃⌥1`–`⌃⌥9`
space shortcuts start working from the keypad on upgrade, with no
configuration — a bigger win than the resize case that prompted
this.

**3. One home for the twin relation.** The equivalence lives in
exactly one map, read by both hotkey registration and
`KeyboardMatrix`'s highlighting — never a copy in each.

**4. `KeyboardMatrix` draws the keypad.** Its docstring already
sets the bar: *"a key drawn as free that `parse` cannot name is a
key the user is invited to bind and then cannot."* Today no
keypad is drawn, so this is a clean absence rather than a broken
promise — but adding the codes without the board would create
exactly that mismatch in reverse. The ANSI/ISO difference is
already modelled as data (`PhysicalType`), so the keypad is a
table, not a conditional.

### Alternatives considered

**Ten distinct keypad keys, no aliasing.** Consistent with the
existing bind-by-physical-position model and needs no equivalence
relation at all — conflict detection, the reverse
`keyName(for:)` map and matrix highlighting would all work
unchanged. Rejected because it gives existing users nothing on
upgrade, surprises people who expect the keypad to work like the
number row, and would force any digit-keyed default to seed eight
rows instead of four.

**The accepted cost of aliasing**, recorded deliberately rather
than discovered later: ten keys a power user might have wanted to
bind separately are given up permanently. Judged the right trade
— "the keypad works like the number row" is what people expect,
and surprising them to preserve a niche capability is backwards.

### Extra context

**Num Lock.** Apple keypads have no Num Lock (a Clear key sits
where PC keyboards put it), so they always send keypad codes. On
a third-party PC keyboard with Num Lock *off* the keypad sends
navigation keys instead, so this works whenever Num Lock is on —
its normal state.

**Why it matters beyond convenience.** The keypad is the only key
family that is identical in both physical position *and* printed
legend on every layout on earth. The number row is position-
stable but not legend-stable: on AZERTY the digits are the shift
layer, so `ComboSymbols` (which resolves legends through
`UCKeyTranslate`) renders `⌥⌘'` where a US user sees `⌥⌘4`.

hajiboy95:
## Description

Makes the numeric keypad bindable. `KeyCombo.keyCodes` stopped at
the number row, so a keypad press matched no binding and did
nothing — for every verb, not just the resize case that surfaced
this.

**The keypad's ten digits are now the same key as their
number-row twin.** A binding written `4` fires from either
physical key, so every existing `⌃⌥1`–`⌃⌥9` space shortcut starts
working from the keypad on upgrade with no configuration. Every
other keypad key (`+ − × ÷ . = enter clear`) is its own bindable
key, and `keypad0`–`keypad9` are accepted as spellings of the
plain digits.

Aliasing follows the platform rather than our own model: AppKit
menu key-equivalents match on the CHARACTER, so `⌘1` and
`⌘`+keypad-`1` are one shortcut in essentially every Mac app.
KiwiDesk binds by key code for international position-stability,
not to tell a keypad from a number row.

`KeypadKeys` is the one home for the relation. Registration reads
it to claim the twin as a second physical key beside the authored
one; `KeyCombo.keyName` reads it to canonicalise a captured keypad
press back to its digit, so the recorder round-trips and writes
`4` whichever key you pressed.

## Related Issue(s)

Closes #1074

## Checklist

  (see AGENTS.md §3)
  required)
  passes

## How I verified

**Not exercised in the running app — and I could not.** This
binds physical keys I cannot press: confirming it needs someone
at a keyboard with a numeric keypad. The first box is deliberately
left unchecked rather than checked loosely.

What that check should be, on a full-size keyboard:

1. `⌃⌥` + keypad `1` switches to the first space, exactly as
   `⌃⌥1` on the number row does — with no config change and no
   new row in the Keybindings editor.
2. In the Settings recorder, pressing `⌃⌥` + keypad `4` records
   the row as `control+option+4`, not as a keypad spelling.
3. Binding `keypadenter` in Lua fires from the keypad's enter
   key and from nothing else.
4. On a third-party PC keyboard, Num Lock **off** stops the
   keypad digits firing (it sends navigation keys instead) —
   expected, and documented.

What I did verify:

- `swift build` clean; full suite green (4186 tests, 793 suites);
  `./scripts/lint.sh` exits 0 with zero errors.
- 12 new tests in `KeypadKeysTests`, and each was **proved to
  red** by mutating the code it watches — transposing the twin
  table, dropping the twin registration, dropping the `keyName`
  canonicalisation, reporting a twin refusal as a failure, and
  dropping the display-name pin. All five went red; the tree was
  restored and re-run green.

## Notes for Reviewer

**Two rulings worth agreeing or disagreeing with explicitly:**

1. **The aliased set is closed to the ten digits, on purpose.**
   The tempting generalisation — "a keypad key mirrors whatever
   character it prints" — needs an exception the moment it is
   written, because keypad `+` has no number-row twin at all
   (main-block `+` is `⇧=`). The accepted cost: ten keys a power
   user might have wanted bound separately are gone for good.
   Argued in `docs/design-decisions.md`.

2. **The keypad is deliberately NOT drawn on the Settings
   board.** `KeyboardMatrix.PhysicalType` knows only ANSI/ISO/JIS
   and macOS exposes no "a keypad is attached" signal, so drawing
   one would show every laptop a block of keys it does not have.
   Bindable-but-undrawn is already the board's shape — `f1`–`f12`,
   `home`, `end`, `pageup` and `pagedown` are all bindable and
   none is drawn, and `drawnCodes` states that a key the board
   omits is not counted as free. This narrows what #1074's body
   proposed, and says so rather than quietly dropping it.

**One refactor rides along, forced rather than chosen.**
`KeybindingManager.swift` was at 341 of its 350-line ceiling, so
the activation half moved to `KeybindingManager+Activation.swift`
at the file's existing `// MARK: - Hotkey activation` seam
(341 → 292) — the same split `KeybindingManager+Repeat.swift`
already took. That required loosening `layers`, `suspended`,
`registrar`, `activationFailures` and `activeBindings`'s setter
from `private` to module-internal; the GUI target still cannot
see any of them. If you'd rather that landed as its own PR, say
so and I'll split it.

**Follow-up, not in scope here:** #1075 moves resize onto a
dedicated `⌥⌘` size layer keyed by `4`/`5` and `7`/`8`, where the
keypad's 2×2 block (`7`/`8` directly above `4`/`5`) makes the
axis mapping self-evident. It works on the number row alone, so
it does not depend on this.

https://claude.ai/code/session_01YQr5BwZnhrSuFYUQD69nex


hajiboy95:
Ran the `review-change` pair. Between them: 5 major, 11 minor. Pushed as `10b92bb2`.

**The one worth reading.** `rowSidePinnedToKeyCodes` could not fail. It asserted `rowTwin(of: keypadTwin(of: row)) == row` — true by construction of the dictionary for *any* table, transposed or not, i.e. precisely the swap its own docstring claimed to catch. My original mutation proof passed it only because I'd written a *half*-transpose that orphaned a row code; a true swap of the 4 and 5 rows left it green. It now pins the pad↔row pairing both ways against each digit's own Carbon constant, and I re-proved it with the true swap.

**A real defect it surfaced.** Keypad Clear prints no character on any layout, so it fell through to the uppercased key *name* — a literal `KEYPADCLEAR` inside a compact chord in the recorder, the Shortcuts list and the menu. Fixed with `⌧`, and the two non-printing keypad keys now carry a guard.

**Also fixed:** six of the eight non-digit key codes were unpinned literals (now held against `kVK_ANSI_Keypad*`); the twin is registered only once the *authored* key has landed, so the report and the keyboard can't disagree; `layers` and `suspended` go back to `private` (read via `bindings(for:)` / `isSuspended`) — `layers` owns the Lua registry refs, and an internal setter dropped the compiler's enforcement for no gain; `KeypadKeys` drops the sorted-order legend derivation for one three-column table, and `names`/`displayOverrides` become `static let` instead of rebuilding two dictionaries per `keyName` call.

**Two residues I documented rather than fixed**, both flagged by review and both accepted deliberately:

- `KeyboardCensus.takenKeyCount` counts codes the board doesn't draw, so binding `⌃⌥keypadplus` raises "Keys taken" while lighting no cap. Already true of `f1`–`f12` and the navigation keys; widened by eight codes here.
- A printing keypad key draws the same glyph as its main-block twin — `⌃⌥keypaddecimal` and `⌃⌥period` both render `⌃⌥.`. The bindings stay distinct (conflict detection compares key codes), so it's a display collision only. Left alone because macOS offers no keypad-distinguishing glyph and a `Num`-style English prefix minted in Core would be neither localizable nor Apple-native. This one gets an `accepted-limitations` row.

**One open question from review I did not decide unilaterally:** whether the glyph collision deserves a proper fix later. Say the word and I'll file it.

Gate re-run: full suite green (4190 tests, 793 suites), `lint.sh` exits 0, and all five new/changed assertions re-proved red under mutation.

Issue #1056, opened 2026-08-27
💡 Resize: hold the chord to keep resizing until release


hajiboy95:
### Use case

Resizing by keyboard is one step per press. Getting a window
from its current size to the one you want means pressing the
chord over and over — the more so on a scrolling slot, whose
`resize.step` moves it a fixed number of points regardless of
how far you have to go.

Every other keyboard adjustment on a Mac repeats while you hold
it: arrow keys, volume, brightness. Resize is the one place
KiwiDesk asks for a drum roll instead.

### Proposed solution

Hold the resize chord and have it keep applying until release —
an initial delay, then a repeat interval, the ordinary key-repeat
shape.

Worth deciding as part of it: whether the repeat should
**accelerate** while held (coarse adjustment gets fast, and the
first press stays precise), and whether the same treatment
belongs on any other repeatable verb — `focus` and `swap` are
the obvious neighbours, and the answer may be no for those,
since overshooting focus is more annoying than overshooting a
size.

### Alternatives considered

Raising `resize.step` so each press does more. It trades away
precision for reach and does not actually solve it: you then
cannot land on the size you want.

Mouse edge-drag already gives continuous resize, so the gap is
keyboard-only — but the mouse is exactly what someone driving a
tiling manager by keyboard is trying not to reach for.

### Extra context

**What ships today.** `CarbonHotkey` installs a handler for
`kEventHotKeyPressed` only — no release handler and no repeat
timer — and a Carbon-registered hot key does not auto-repeat on
its own. So one press is strictly one step, by construction
rather than by oversight.

**The design question, and how far it actually reaches.**
`docs/accepted-limitations.md` records that holding a key to
resize a **floating** window under-accumulates: that path
re-bases each step on `state.windows[id].frame`, which is
echo-fed from AX, and mid-animation the echo lags, so a step
issued before the previous one settled reads stale geometry.

That limitation does **not** reach tiled resize, and the
distinction decides most of this issue's size. A tiled resize
never re-reads geometry: `writeCappedScrollSlot` takes the
STORED `slotSize`, adds the delta and writes it back, and the
BSP and stack paths do the same with their resolved ratio. That
is already accumulation against the commanded value — which is
precisely the model a held key needs. So for tiled windows,
including scrolling, repeat should accumulate correctly with no
change to how the delta is applied.

The floating path is the one that needs a decision, and the
options are ordinary: accumulate against the commanded target
instead of the observed frame, or coalesce a held run into one
[...]

Issue #270, opened 2026-07-15
💡 Keyboard-layout ergonomics for non-US layouts (defaults + docs + Lua recipe)


hajiboy95:
## Context

KiwiDesk binds keys by **physical key position** via a hardcoded US-QWERTY keycode table (`KeyCombo.keyCodes`), verified end-to-end:

- **Registration** (`KeyCombo.parse` → Carbon `RegisterEventHotKey`, `KeybindingManager.swift:252`, `CarbonHotkey.swift:99`): no layout translation — a combo fires on the physical US key position.
- **Recording** (`ChordRecorder`): captures the physical `event.keyCode`, relabels with its US name.
- **Display** (`ComboSymbols` + `LayoutKeyGlyph`, #23): the *only* layout-aware stage — shows the current layout's glyph at that physical position.
- Nothing stores the keyboard layout/language.

**Symptom (German user):** the default `option+minus` (resize) fires on the physical US `-` position — the `ß` key on QWERTZ — and displays as `⌥ß`. Hand-authoring config requires knowing US physical names (`-` is `"slash"` on QWERTZ).

## Decision (architect + designer consult)

**Keep the physical-US model.** It's the tiling-WM genre norm (skhd/yabai/AeroSpace), preserves roaming-profile muscle memory, and keeps `KeyCombo.parse` pure/testable. **Rejected:** character-based registration, auto per-key "remap only the inaccessible" (usability trap + unsolved bin-packing; on macOS AltGr *is* Option, so punctuation defaults are frequently "inaccessible"), and storing the layout. The pain is mostly a **default-choice** problem, not a model problem.

## Scope

1. **Defaults audit (code).** Re-pick shipped default bindings toward **letters / digits / arrows** (stable label *and* position across Latin layouts) and away from roaming **punctuation** (`minus`/`equal`/brackets). Seam: default binding tables only (`DefaultKeybindings.swift`, `LuaBindingBody.swift`) — does not touch `KeyCombo`. Start by listing current defaults and flagging which roam on non-US layouts.
2. **Docs (reference/guide).** Name the model plainly: bindings are *physical keys by US label* (skhd/AeroSpace convention); the **recorder is the authoring path** (press the key you want — no US knowledge needed); display shows your layout's glyph.
[...]

PR #1080, opened 2026-08-28
feat(keys): move resize to a dedicated ⌥⌘ size layer (#1075)

### LINKED ISSUE #1075: 💡 Move resize to a dedicated ⌥⌘ size layer: ⌘ means two things, and the one held verb is on the heaviest chord

hajiboy95:
## Description

Resize sat on tier 3 of the `⌃⌥` ladder, which made `⌘` mean two
unrelated things at once — "resize" with an arrow, "and follow"
with a digit. The ladder escalates over one idea (who the verb
acts on) and resize never belonged to it: focus, swap, move and
follow all place a window in the flat array, while resize changes
a weight or a scroll-slot domain.

Size now takes a base of its own:

| | shrink | grow |
|---|---|---|
| **width** | `⌥⌘1` | `⌥⌘2` |
| **height** | `⌥⌘4` | `⌥⌘5` |

`⌘` on the ladder means exactly one thing again, and the
positional scheme reads as one rule with a separate "`⌥⌘` = size"
beside it — simpler than the three-rung version, not just
different.

It also fixes an inverted ergonomic: #1056 made resize the one
verb a user *holds*, and it sat on the heaviest chord in the
scheme — three modifiers with `⌃` and `⌘` at opposite ends of the
left cluster. `⌥⌘` is a single thumb roll.

Digits rather than arrows because an arrow reads two ways on a
tiled window — "which axis and sign" and "which way the edge
moves" — and which edge is free depends on where the window sits
in the array, so the same arrow grows a right-column window and
shrinks a left-column one. Within a pair the higher digit grows. `1`/`2` and `4`/`5` form a
2×2 block on a keypad, `4`/`5` directly above `1`/`2`, and on the
number row `3` separates the pairs so a mistimed reach for one
axis can't land on the other.

The digits are **measured, not assumed**: `⌥⌘8` is macOS's
Accessibility Zoom toggle (`com.apple.symbolichotkeys` id 15,
read 2026-08-28), which an earlier draft of this layer had taken
for Grow height. `SystemShortcuts.map` now carries the ⌥⌘ family
so the app warns, and the seed is held against it.

## Related Issue(s)

Closes #1075

## Checklist

  (see AGENTS.md §3)
  required)
  passes

## How I verified

**Not exercised in the running app — and I could not.** Seeing
the new seed requires a config with no bindings anywhere (the
#270 guard), and confirming the chords requires pressing them.
The first box is left unchecked rather than checked loosely.

What that check should be, on a fresh config:

1. `⌥⌘2` grows the focused window's width, `⌥⌘1` shrinks it;
   `⌥⌘5` / `⌥⌘4` do the same for height.
2. Holding `⌥⌘2` keeps resizing and accelerates (#1056), and
   releasing stops it — this is the ergonomic the move is for.
3. `⌃⌥⌘←`/`→`/`↑`/`↓` now do nothing on a fresh install.
4. An **existing** install is untouched: `⌃⌥⌘`+arrows still
   resizes, and `⌃⌥K` still shows those chords.
5. `⌥⌘` doesn't steal anything: browser tab switching
   (`⌥⌘←`/`→`), Force Quit, Dock hiding and Finder search all
   still work, and so does Zoom (`⌥⌘8` / `⌥⌘-` / `⌥⌘=`) if you
   have its keyboard shortcuts enabled.

What I did verify:

- Full suite green (4180 tests, 794 suites); `lint.sh` exits 0.
  The seed↔catalog byte-for-byte parity guard
  (`DefaultSeedCatalogParityTests`) is unaffected — the GUI
  catalog authors the Lua and labels, and the combos live only in
  `DefaultKeybindings`.
- 6 tests in `SizeLayerSeedTests`, each **proved to red** by
  mutation: putting a size row back on an arrow, seeding a
  positional verb onto `⌥⌘`, and swapping grow/shrink within the
  width pair. Tree restored and re-run green.

## Notes for Reviewer

**This amends a recorded ruling, and the amendment is the part to
check.** `docs/design-decisions.md` rejected `⌘⌥` outright over
"always-on system shortcuts (Force Quit, Dock, Hide/Minimize)".
Every collision it named is a letter, an arrow or `esc` —
`⌥⌘esc`, `⌥⌘D`, `⌥⌘H`, `⌥⌘M` — and none is a digit; the same
passage also establishes the base is text-safe, since composition
happens only when the modifier is exactly `⌥` or `⌥⇧`. So the
narrow use survives the original objection.

The boundary is part of the ruling, and is now an obligation in
`.claude/rules/profiles.md` with the §5 row to match: **`⌥⌘`
carries digits and safe letters only** — never arrows (`⌥⌘←`/`→`
is next/previous tab in Chrome, Safari and Terminal, and a global
Carbon hotkey pre-empts the frontmost app), never `esc`, `space`,
`D`, `H` or `M`.

**No migration, per your call.** The seed guard fires only when
no layer carries a single binding, so every existing config blocks
it, and `digitTopUp` is strictly additive. Nobody is moved.
`ShortcutsReferenceBuilder` renders the *live* layer rather than
the defaults, so an existing user's Shortcuts panel stays
accurate — that was the one thing I'd expected to go stale, and
it doesn't.

The accepted cost, recorded rather than discovered: there's no
restore-to-defaults affordance in the Keybindings editor, so an
existing user who *wants* the new map hand-edits four rows. And
prose that names a chord now has two populations to serve, which
is why the resize workaround in `accepted-limitations.md` points
at the `⌃⌥K` panel instead of naming a chord.

**Depends on nothing.** #1074 (keypad) is independent — this map
works on the number row alone; the keypad just makes the 2×2
mnemonic visible to people who have one. The two branches don't
touch the same files.

https://claude.ai/code/session_01YQr5BwZnhrSuFYUQD69nex


hajiboy95:
Review round done. **The digits changed** — the reviewers caught a blocker I shipped.

## ⌥⌘8 is macOS's Accessibility Zoom toggle

I asserted in the ruling that every ⌥⌘ collision "is a letter, an arrow or `esc`, and none is a digit". That was false, and I'd written it from reputation rather than measurement. Reading `com.apple.symbolichotkeys` on macOS 26.6 (2026-08-28), the chords whose modifiers are exactly ⌥⌘ are:

```
id 15  '8'    keycode 28  enabled=False   Zoom on/off
id 17  '='    keycode 24  enabled=False   Zoom in
id 19  '-'    keycode 27  enabled=False   Zoom out
id 23  '\'    keycode 42  enabled=False
id 52  'd'    keycode  2  enabled=True    Dock hiding
id 65  space  keycode 49  enabled=True    Finder search
```

`⌥⌘8` was **Grow height**. Zoom's shortcuts are gated on Accessibility ▸ Zoom ▸ "Use keyboard shortcuts to zoom" and ship off — which is exactly why a reputation-based list misses them — but for anyone who has turned them on, that row would have been dead or double-firing, and `RegisterEventHotKey` refuses silently. Shipping a default that breaks specifically for people using an accessibility feature isn't a trade worth making.

(It also retroactively settles the `-`/`=` idea from earlier: macOS owns that exact pair for Zoom in/out.)

## New map

| | shrink | grow |
|---|---|---|
| **width** | `⌥⌘1` | `⌥⌘2` |
| **height** | `⌥⌘4` | `⌥⌘5` |

All four are clear of the measured table. They earn the keys twice over: `4`/`5` sit directly above `1`/`2` as a 2×2 block on a keypad, and on the number row `3` separates the pairs, so a mistimed reach for one axis can't land on the other. `{2,3}/{5,6}` was rejected — touch typing splits the row between `5` and `6`, so that pair straddles the hand boundary.

## The rest

- **`SystemShortcuts.map` had no ⌥⌘ entry at all**, so nothing warned a user binding one and the keyboard board drew them free — this branch's own premise went unchecked. It now carries the ⌥⌘ family, and `SizeLayerSeedTests` holds every seeded row against it rather than only against the rest of the seed.
- **`sizeLayerCarriesOnlySize` was the empty-watched-collection shape.** Its loop is skipped when nothing is seeded on ⌥⌘ — precisely the state after the regression it exists to catch — so it would have gone green having looked at nothing. Now asserts the subset is non-empty first, and the mutation that empties the base reds it.
- The redundant arrow test folded into `everySizeRowIsOnTheSizeLayer`; the two couldn't fail apart.
- **Doc comments no longer state #1074's keypad reach in the present tense.** It's unmerged, so those keypad keys fire nothing today. The user guide states only the number-row reason for now and gets the keypad sentence when #1074 lands.
- **The boundary obligation moved to `input-and-animation.md`**, whose `paths:` actually reach the chord authors it binds — `profiles.md` globs `Config/**` and would never have loaded for someone editing `SystemShortcuts.swift`. It's now stated as *check the register*, not as a prose list of macOS chords: a rule that re-enumerates macOS is how this went wrong once already.
- The ruling names the ladder's third class (toggles + app chrome, where `⇧` means "broader scope" — a second sense of `⇧` that predates this change), applies the app-collision criterion to digits honestly instead of assuming past them (⌥⌘+digit is Xcode's inspector family; the line drawn is reach, not kind), records what the two-base scheme costs a user, and annotates #270's rejection sentence where a reader meets it.
- `DefaultKeybindings.swift` hit 326 lines against the 250 target, so the size rows split into `DefaultKeybindings+Size.swift` (seed file back to 258).

Gate re-run: full suite green (4181 tests, 794 suites), `lint.sh` exits 0, and four mutations proved red — the Zoom chord, an emptied size base, the register losing its Zoom entries, and a swapped grow/shrink pair.

Issue #1084, opened 2026-08-28
🐛 Resize presses queue on a starved main actor: the row freezes, then jumps several steps


hajiboy95:
### KiwiDesk version or commit

`1493cedf` (the #1083 branch), signed local build.

### macOS version

macOS 26.6.2

### Mac model type

Apple silicon (owner's development Mac).

### Steps to reproduce

1. Load the machine (a browser burning ~150% CPU across its
   processes is enough — measured with Zen at 98% plus two
   plugin containers).
2. In a **scrolling** space with 5 windows, hold or repeatedly
   press the resize chord.

### Expected vs. actual behavior

**Expected:** the row resizes at a steady cadence — a held key
repeats at the system rate (~90 ms).

**Actual:** the window freezes for up to two seconds, then jumps
several steps at once. Owner report: *"it stucks sometimes in
between"*, reproduced with both a held chord and repeated
clicks.

**Not a refusal.** Every press succeeds. Instrumented run, 96
presses, `refusal=nil` on every one and `write` advancing by
exactly the step each time:

```
scroll-resize: delta=-50 stored=1516 drawn=1516 area=1716 write=1466 refusal=nil
scroll-resize: delta=-50 stored=1466 drawn=1466 area=1716 write=1416 refusal=nil
scroll-resize: delta=-50 stored=1416 drawn=1416 area=1716 write=1366 refusal=nil
```

The defect is entirely in **when** those presses are processed.

### Relevant output and logs

Inter-press intervals across those 96 presses:

```
min 12 ms   median 315 ms   max 2087 ms
under 30 ms (burst flush): 8
over 300 ms (stall):      48
```

A held chord should be a flat ~90 ms. Instead the sequence
stalls — 1143, 919, 1099, 1471, 1246, 1533, 2087 ms — and then
flushes three presses in 12–15 ms. Presses queue on a starved
main actor and are processed in bursts, which is exactly the
"freeze, then jump several steps" the owner sees.

Machine state at the time (the load is real and mostly not
KiwiDesk's):

```
151%  node (dev servers)     98%  Zen      56%  Zen plugin-container
 49%  Claude                 40%  WindowServer     20%  KiwiDesk
```

### Other window managers or AX tools running?

None. KiwiDesk only.

### Extra context

**Suspected mechanism.** Each press runs a full retile and
issues frames for every window in the row — 5 windows of AX
`setFrame` plus animation retargets, per press. Under load those
[...]

Issue #1057, opened 2026-08-27
💡 Resize an oversize slot from what is drawn, not from what is stored


hajiboy95:
### Use case

Set a scrolling slot larger than the screen you are on —
`scroll.set_slot_size(3000)` on a 1512pt display, or a profile
written at the desk and applied on the laptop — and the resize
key goes dead. The layout draws `min(along, …)`, so the slot on
screen is 1512; but a `resize` press counts from the stored
3000, so the first several presses move a number nobody can
see. You press shrink repeatedly, nothing happens, and then it
starts working.

Observed in device QA (2026-08-27, owner) while driving #966's
ceiling: an interactive grow now stops at the screen and a
single shrink press responds immediately, which is exactly
right — and the same slot reached by configuration still has to
be unwound.

### Proposed solution

Seed an interactive resize from what the layout DRAWS rather
than from what is stored, when the two differ.

- A shrink press starts from the drawn size, so the first click
  has visible effect.
- A grow press still refuses, because there is nothing to grow
  into — unchanged.
- Nothing else writes, so the stored value survives everything
  except an explicit interactive resize.

The last point is what makes this the right shape rather than
a smaller one: the configured value keeps travelling between
screens, and is only overwritten at the moment the user
deliberately resizes it on a particular screen — which is
exactly when they are saying "make it different here".

### Alternatives considered

**Cap the value at `set_slot_size` time**, to the axis extent of
the current screen. Simplest, and the first thing that comes to
mind, but it destroys the property the ceiling was designed
around: `set_slot_size` runs from `init.lua` at startup and from
profile applies, often before the display set is final, so
undocking to a laptop would permanently rewrite 3000 to 1512 —
and re-docking would not bring it back. A config value has to
survive travelling between screens; that argument is in
`docs/design-decisions.md` and is why #966 put the ceiling at
the interactive write site instead.

**Clamp the store on the first shrink**, i.e. treat any resize
of an oversize slot as consent to trim it. Equivalent to this
proposal for the shrink case, but it also trims on a grow, which
is the config-destruction bug #966 fixed.

**Leave it.** Defensible — the workaround is to set a sane value
— but the "keys do nothing for a while" symptom is the same
class of defect as #966 itself, one path over.

### Extra context

**Not a regression.** This predates #966's ceiling: before it,
`set_slot_size(3000)` already drew `min(along, …)` while a
resize counted from 3000. The ceiling fixed the interactive-grow
[...]

Issue #1055, opened 2026-08-27
🐛 Learned app size bounds: no ceiling to act on, and scrolling re-learns them on every resize


hajiboy95:
### KiwiDesk version or commit

1.0.1 — branch `fix/scrolling-resize-anchor` @ `c1595e55`
(#966's branch; the two resize clamps there are unrelated to
this, and this reproduces without them).

### macOS version

macOS 26.6.2 (25G83)

### Mac model type

- [x] Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

Owner device QA, 2026-08-27, on a Scrolling space with two
constrained apps. Either app shows it:

**A — an app with a MAXIMUM width (Apple's Settings; the
Android emulator behaves the same way, growing then snapping
back).**

1. A Scrolling space, horizontal, with Settings and at least
   one other window in the row.
2. Focus Settings and press `resize` to grow the slot.
3. Watch the neighbour: it slides fully aside as the row is
   laid out for the full asked width, then Settings snaps back
   to its own maximum and the row re-packs.

**B — the learned bound going stale when a DIFFERENT window is
resized.**

1. Same space. Let the Android emulator settle, so its
   grow-and-clamp-back dance has finished once.
2. Focus another window and resize it a few times.
3. The emulator runs the whole dance again — it grows, clamps
   back, and has to re-learn what it already knew.

**C — no refusal at a MINIMUM in scrolling.**

1. Settings also refuses to go below its own width.
2. Focus it and shrink repeatedly.
3. Nothing happens, and nothing says why.

### Expected vs. actual behavior

**Expected.** A window whose app enforces a size is remembered
in BOTH directions, and neither direction re-learns for free:

- Resizing the constrained window itself, in either direction,
  does nothing visible and says so — the size-limit cue #933
  already shows on the BSP/stack paths.
- Resizing a DIFFERENT window changes every window except the
  constrained one, which keeps the size its app insists on.
- Nothing re-runs the grow-and-clamp dance once it has been
  learned.

**Actual.** Only the minimum direction is remembered in a form
anything can act on, the memory goes stale whenever another
window is resized, and the maximum direction produces the visible
overshoot in A with no cue in C.

### Relevant output and logs

Not captured — reported from live use. The mechanism is legible
in the source and stated under Extra context; a log capture of
the learn dance would show repeated `size bound candidate for
window <id> (resize echo); probing once more` lines for the same
window.

### Other window managers or AX tools running?

[...]

Issue #1049, opened 2026-08-26
🐛 Android emulator sends tiling and focus berserk: transient resize compliance wipes the #677 ladder; toolbar raise provokes an endless focus steal


hajiboy95:
### KiwiDesk version or commit

1.0.1 (main at 85a033f9, dev build from `.build/app`)

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Start the Android emulator windowed
   (`emulator -avd Pixel_8_API_36`) on a scrolling space with
   other windows, so its slot is wider than the phone's aspect
   ratio allows (slot ~1626×1005 on a 16" screen).
2. Let KiwiDesk tile it, or force a retile
   (`kiwidesk set_gap_global 8`).
3. Hands off. Watch the window and the unified log.
4. To end the fight, pick the window up and place it by hand;
   hiding the emulator instead brings it back on its own.

### Expected vs. actual behavior

Expected: the #677 size-bound learner notices the emulator will
not hold the slot size, confirms the refusal after two probes,
and the layout re-packs around the window's own size. Focus
stays where the user put it.

Actual — two coupled loops, both confirmed hands-off:

**Resize berserk.** The emulator *transiently complies*: it
animates out to the full asked size (1626×1005), holds it
~0.4 s, then snaps itself back to 439×1004 (device aspect
ratio) in one jump. The compliance echo runs the learner's
`complied` sweep, which clears the candidate ladder — so the
snap-back re-seeds a fresh candidate every cycle and "twice in
a row" never accumulates. `size bound candidate for window …;
probing once more` fires every ~0.85 s indefinitely (22+ cycles
captured), each one a retile, each retile a new expand-and-snap
dance on screen.

**Focus stealing.** After every honored focus, the float raise
re-raises the emulator's floating toolbar (w227844). Qt answers
an AXRaise with app activation and keys its MAIN window
(w227843) — a window the raise never stamped, so the per-window
z-order echo net cannot see it, `handleWindowFocused` honors
the steal (`selfEcho=false`), warps the pointer there, a
restore re-asserts the previous focus, the float raise fires
again… a closed loop with a ~2 s period and zero user input.
Hiding the emulator does not stick: it re-activates itself and
is honored again on re-track.

### Relevant output and logs

Frame poll (CGWindowList, 20 ms) during one cycle — full
compliance, hold, one-step snap back:

```
[...]

PR #1085, opened 2026-08-28
fix(tiling): a learned bound informs the layout, never refuses a press (#1083)

### LINKED ISSUE #1083: 🐛 Redraw latency is recorded as a size bound, so resize refuses at limits that do not exist

1.0.1, `afe3fa40` (current `main`), signed local build.

### macOS version

macOS 26.6.2

### Mac model type

Apple silicon (owner's development Mac).

### Steps to reproduce

1. Put a space in **scrolling** mode with several windows, at
   least one of them a slow-redraw app (reproduced with **Zen**).
2. Focus a window that is nowhere near any size limit.
3. Press the resize shrink chord several times in a row, at
   ordinary typing speed (~1/second is enough — no hold needed).
4. Now try to resize that window further, or try to grow it.

**Expected:** the window keeps resizing until it reaches a real
limit — the configured `min_window_size`, the viewport, or a
size its app genuinely refuses.

**Actual:** resizing stops early and the size-limit pill claims a
limit that is not real — "min size" on a window visibly larger
than the minimum, and on Zen a "max size" refusal with room left
on screen. Dragging the same window's edge **by hand still
works**, which proves the app itself imposes no such limit.

### Relevant output and logs

Eight minutes of ordinary use produced **16 bound confirmations
against a single candidate** — the ladder is supposed to need a
candidate and then a confirming re-encounter, so nearly every
confirmation is skipping the ladder:

```
  9  size bound confirmed for window N at NxN (resize echo)
  7  size bound confirmed during retile
  1  size bound candidate for window N; probing once more
  1  size bound staled by a genuine resize
```

The confirmed values are the giveaway — one window, one press per
second, each "bound" exactly 30 pt above the last:

```
21:28:47.432  size bound confirmed for window 347220 at 984x1011
21:28:48.409  size bound confirmed for window 347220 at 954x1011
21:28:49.126  size bound confirmed for window 347220 at 924x1011
21:28:49.868  size bound confirmed for window 347220 at 894x1011
21:28:50.436  size bound confirmed for window 347220 at 836x1011
```

30 pt is the resize step. Each recorded "bound" is the window's
size *before* that press. The heights are all `1011` — the slot
height, not any app's limit. Two different windows also confirmed
the identical bound 44 ms apart (`145372` and `347220`, both at
`1127x1011`), which is what a shared drawn geometry looks like,
not two apps independently refusing.

None. KiwiDesk only.

### Extra context

**Suspected mechanism** (`SizeBoundLearner.observeAxis`, the
baseline arm):

```swift
if let baseline, baseline > 0,
   EffectiveSizeBound.matches(baseline, current),
   !candidateEntries.contains(where: {
       EffectiveSizeBound.matches($0.asked, asked)
   })
{ return promote(id, asked: asked, answered: current, axis: axis) }
```

The arm exists so that a pre-ask size *unchanged through a whole
animation of size-sets* can complete "the same answer twice"
without a second probe. But it never consults `settledRead`, and
on the **echo** channel a reading equal to the baseline is
ambiguous: it means either "the app refused" or "the app has not
redrawn yet". A slow-redraw app answers the second way, so the
learner records latency as a refusal and mints a bound at the
size the user was already at.

That false bound then feeds `ScrollSlotDomain`, whose `appMin` /
`appMax` arms refuse in place and cue the pill — which is why the
sentence names a limit the window is nowhere near.

`SizeAnswerChannel.isSettledRead` already exists and is described
as the owner of the settled/raw verdict (#1049); the baseline arm
appears to be the one promotion path that does not ask it.

**Why now.** The trade is documented in the neighbouring
comply-then-revoke arm — *"a USER resize landing inside the ask's
echo grace right after the compliance confirms a false entry"* —
and was sized for isolated presses. Anything that issues asks
faster than a window settles turns it from rare into routine, so
this wants fixing before #1082 lands (a held glide asks once per
display frame, ~16 ms apart).

Recoverable, not destructive: a genuine resize or a settled
compliance clears the false entry, which is why the window frees
up again later.

hajiboy95:
Resize refused at limits that did not exist. Two commits: the
learner stops inventing them, and a learned bound stops being
allowed to veto a press at all.

fixes #1083

## What was happening

A "size bound" means *the app refused to be sized past here*.
The learner was minting them from redraw latency instead.
Device capture (2026-08-28, macOS 26.6.2, load average 9.7):

```
21:28:47.432  size bound confirmed for window 347220 at 984x1011
21:28:48.409  size bound confirmed for window 347220 at 954x1011
21:28:49.126  size bound confirmed for window 347220 at 924x1011
21:28:49.868  size bound confirmed for window 347220 at 894x1011
```

30 pt is the resize step. Each "bound" is the window's size
*before* that press, and every height is `1011` — the slot's,
not any app's. Sixteen confirmations against **one** candidate
in eight minutes, where the ladder is supposed to need two
observations.

Those bounds fed `ScrollSlotDomain`, which refused the next
press and cued a pill naming a limit the window was nowhere
near. A window resized both ways ended pinned between a false
minimum and a false maximum. Dragging the edge by hand worked —
which proved the app imposed nothing.

## 1. Only a settled read may confirm

On the echo channel, a reading equal to the pre-ask baseline is
ambiguous: *"the app refused"* and *"the app has not redrawn
yet"* are the identical frame. Under load the second is ordinary
for **any** app — this was first reproduced on Ghostty, which
our own docs cite as the fast one.

The baseline arm and the two-observations ladder now require the
settle probe, the read that waits out
`sizeBoundProbeGraceSeconds`. Raw echoes still seed, refresh and
clear. #1049's comply-then-revoke arm is deliberately untouched:
it fires only after an echo reported the window AT the asked
size, which is positive proof it took it.

Cost: a genuine limit is learned one probe grace (~0.6 s) after
its animation settles, instead of at echo time.

## 2. A guess informs the layout; only a fact may refuse a press

Hardening detection was a losing game — three separate paths can
mistake latency for a refusal, each guarded by a different
heuristic about whether the app has answered, and all of them
degrade together under load. The durable fix is that a wrong
guess must not be able to block anything.

`effectiveMinSize` answers the configured `min_window_size`
alone; `effectiveMaxSize` answers nil; the scrolling caller
feeds the domain no learned limit and no longer measures the
press from the learned consume.

The asymmetry decides it: wrong permissively costs a window that
does not fill its region — the accepted split-layout residue,
self-correcting on the next retile. Wrong restrictively costs
the user the feature and states a falsehood while doing it.

`ScrollSlotDomain`'s own-minimum and own-maximum arms are
**kept**: correct for a limit that is KNOWN, and a trustworthy
source would plug into them unchanged.

## The two floors that moved rather than disappeared

Found by a failing test, not by inspection. The layout draws a
bound-pinned window at its learned limit, and a press measures
from that DRAWN span (#1057) — so with the bound no longer
refusing, a grow wrote that span back into the shared store and
**trimmed the row for every neighbour** (a ~1160 pt auto slot
rewritten to 765). The shrink mirror raised a 300 pt store to
775.

`ScrollSlotDomain` now holds the shared store from both sides —
a grow never reduces it, a shrink never raises it, both
wordlessly.

## Verification

- 4178 tests, lint, release build, site build — green
- `echoAtBaselineNeverConfirms` self-proved: reverting the
  `settledRead` term reds it with the exact symptom
  (`asked: 980.0, answered: 715.0`)
- On device, the same eight-minute window of use went from **16
  bound confirmations to zero** (2 candidates, 0 confirmations)

14 tests changed meaning because the ruling changed. Each was
re-pointed at what still holds rather than deleted: the
split-layout suites now assert the press travels *past* the
learned bound and stops at the configured floor; the ceiling
suite keeps its never-reduce property and drops the refusal; the
neighbour suites keep the #435 cue anchoring for the configured
floor.

## Docs

`design-decisions.md` carries the ruling with the frequency
evidence. `accepted-limitations.md` records that interactive
resizing has REJOINED the app-minimum row. `state-and-layout.md`
and the #1055 entry are amended.

One earlier ruling is superseded and called out in place:
*"a window that cannot follow never resizes the row"* (owner,
2026-08-28 morning) still holds for a limit that is KNOWN, and
the domain still implements it — but a learned bound is not one,
so the caller no longer hands it one.

## Not included

- [#1084](https://github.com/KiwiCanopy/KiwiDesk/issues/1084) — presses queue on a starved main actor (freeze,
  then jump). Measured during verification, unrelated to this:
  every press succeeded, `refusal=nil`, median interval 315 ms
  against a max of 2087 ms.
- The pill still fires *before* the fact for configured limits.
  Making it an explanation *after* the fact is the natural next
  step and is not attempted here.


hajiboy95:
**Device verification (owner, 2026-08-28, macOS 26.6.2)** — the
branch as it now stands, signed local build:

> this size bounding works really good! sometimes windows get
> below but that's the tradeoff, it's working reliable

Instrumented evidence across the session, same machine, same
kind of use:

| | before | after |
|---|---|---|
| bound confirmations / 8 min | 16 | 0 false |
| candidates behind them | 1 | every confirmation |
| confirmed values | 984, 954, 924, 894 — the window's own pre-press widths, one resize step apart | 500, 400, 825 — plausible app minimums, each after a candidate, all via the settle probe |

The "sometimes gets below" is the expected residue rather than a
regression: a bound must be LEARNED before it can bind, so a
first encounter with a real minimum can overshoot once and the
ledger holds it afterwards. That case is already carried by
`docs/accepted-limitations.md` ("a bound not yet learned").

**Scope note for reviewers.** This PR no longer contains the
permissive ruling that earlier commits on the branch
introduced — learned bounds refusing a press was implemented,
measured, and reverted on the owner's ruling once the learner
was fixed and its bounds became trustworthy. What remains is the
learning fix, the forward-base rule and their tests.
`docs/design-decisions.md` keeps the permissive alternative and
the frequency argument for it on record, so a future latency
path has the case already written.

PR #1188, opened 2026-08-31
feat(layout): default the BSP split strategy to alternating

### LINKED ISSUE #1181: 💡 Default the BSP split strategy to alternating
### Use case

Out of the box, BSP should behave the way people who reach for a
BSP layout expect it to. Today the default split strategy is
`longest_side` (cut the longer side, keeping windows square-ish);
most tiling-WM users coming from bsp-style managers expect the
classic **alternating** split (horizontal / vertical by depth) —
that alternation *is* the mental model "BSP" carries. A new user
opening windows under the default reads the longest-side
placement as the layout misbehaving rather than as a policy
choice.

### Proposed solution

Flip the default: `BspParams.strategy` defaults to
`.alternating` instead of `.longestSide` (owner ruling,
2026-08-31 — "that's more what people would expect from bsp").

Scope notes from the dev machine:

- Two default sites move together:
  `Sources/KiwiDeskCore/Layouts/LayoutParams.swift:25` (property
  default) and the sparse-decode fallback at
  `LayoutParams.swift:52` (`decodeIfPresent … ?? .longestSide`).
- No value rename, so no `ConfigMigration` crossing (§5 governs
  renamed *stored values*; both spellings live on unchanged).
  `BspParams.encode` writes `strategy` unconditionally, so every
  GUI-saved config/profile already pins its strategy and is
  unaffected — the flip reaches fresh installs and configs that
  never set the key. That behavior change on update is
  deliberate and gets a `## Highlights` line at release.
- Docs move in the same change set: `docs/lua-reference.md`
  (`bsp.set_strategy`, ~704, and the settings JSON example
  ~4460 if it shows the default), `docs/user-guide.md` ~932,
  `docs/cli.md` ~308 — wherever the default is stated, per
  docs.md parity.
- Tests: any fixture reasoning from the current default pins it
  per tests.md (#660) rather than inheriting it — check
  `LayoutBspTests`, `BspOverrideTests`, the Layout Defaults
  schematic suites, and `SettingsValueReadout` /
  `SettingsDraftDiff` expectations for an implicit
  `longest_side` assumption.
- Starter/preset check: confirm whether `StarterTuning` /
  `StandardProfiles` set the strategy explicitly; if they do,
  rule whether they follow the new default or keep a deliberate
  pin.

### Alternatives considered

- Keep `longest_side` and rely on the Settings picker /
  `bsp.set_strategy` — rejected: defaults are the product's
  opinion, and this default reads as wrong to the audience BSP
  is for. Both strategies remain available; only the default
  moves.
- A per-screen-shape default (like the starter setup's derived
  layouts) — over-engineered for a split policy; one honest
  default plus the existing per-space override covers it.

### Extra context

Filed from the dev machine (owner request). Slotted into the
1.1.2 milestone as an Opus lane: two-line default flip plus docs
and test-pin sweep fits the release's small-and-safe bar.

hajiboy95:
## Description

Out of the box, BSP now **alternates** — horizontal then vertical
by depth — instead of cutting each region's longest side. That
alternation *is* the mental model the word "BSP" carries for the
people who reach for a BSP layout; under `longest_side` a new
user reads the placement as the layout misbehaving rather than as
a policy choice.

Both strategies stay available; only the default moved.
`bsp.set_strategy` and the per-space override are unchanged.

**This changes behaviour on update, deliberately.**
`BspParams.encode` writes `strategy` unconditionally, so every
GUI-saved config and profile already pins its own value and is
untouched — what moves is fresh installs and any config that
never set the key. It gets its own `## Highlights` line at
release rather than arriving silently.

## Related Issue(s)

Closes #1181

## Checklist

  (see AGENTS.md §3)
  required)

## How I verified

`swift build`, `swift test -q` (4304 tests), `scripts/lint.sh` —
green. Release build skipped locally (no concurrency or
`Sendable` change); CI's job runs it.

**No new test for the default itself, deliberately** — the
invariant worth guarding is that the *two* default sites agree
(the property default and the sparse-decode fallback), because
flipping one alone makes a config that omits `strategy` decode
differently from a fresh `BspParams()`. That is already held by
`LayoutParamsCodingParityTests.missingKeysUseDefaults`; I mutated
one site alone and confirmed it reds. A test pinning the value
itself would red on every deliberate retune and catch no
regression (`tests.md`).

Confirmed by grep rather than assumption: neither `StarterTuning`
nor `StandardProfiles` pins the strategy, so both follow the new
default.

## Notes for Reviewer

`review-change` ran. Its most valuable finding is one I would not
have caught:

**The flip silently emptied `LayoutBspTests.longestSideRecursion`.**
On that fixture's bounds both strategies produce identical frames,
so the test kept passing while running `.alternating` under a name
and a comment claiming longest-side recursion — and after the flip
**no test in the tree exercised `.longestSide` past depth 0**.
Pinned, along with `ratioIndependence`, which was the third
unpinned fixture.

Also addressed: the docstring was over the §2.8 budget and
carried the barred kinds (design history, reviewer-facing
justification), so the argument moved to
`docs/design-decisions.md` where a contributor would look before
undoing it. And the marketing site's `layout_bsp` blurb plus the
README both described the *non-default* strategy verbatim
("square-ish splits") — now rewritten to say what the layout is
rather than which strategy it happens to run, so the next retune
leaves them true.

Two fixtures were re-pinned rather than re-derived
(`splitRatioV` wants longest-side because it is about a portrait
screen's shape; `strategyOverride` needs its global pinned *away*
from the value the override writes, or it cannot tell "only that
space" from "every space"). A `guard-prover` run on the repointed
fixtures is in flight; I will report anything it turns up before
this merges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ that list:
 
 | Layout | Behavior |
 |---|---|
-| `bsp` | Binary space partitioning (square-ish splits) |
+| `bsp` | Binary space partitioning (each new window halves a pane) |
 | `stack` | Master zone + stack column (`master_count`) |
 | `scrolling` | Niri/PaperWM-style scrolling columns or rows |
 | `monocle` | Focused window maximized, rest behind it |
--- a/Sources/KiwiDeskCore/Layouts/LayoutParams.swift
+++ b/Sources/KiwiDeskCore/Layouts/LayoutParams.swift
@@ -22,7 +22,9 @@ public struct BspParams: Sendable, Equatable, Codable {
         case alternating
     }
 
-    public var strategy: Strategy = .longestSide
+    /// How each region is cut. Defaults to `.alternating`
+    /// (#1181).
+    public var strategy: Strategy = .alternating
     /// Ratio of side-by-side splits (#56).
     public var splitRatioH: Double = 0.5
     /// Ratio of stacked top/bottom splits (#56).
@@ -49,7 +51,7 @@ public struct BspParams: Sendable, Equatable, Codable {
             try container.decodeIfPresent(
                 Strategy.self,
                 forKey: .strategy
-            ) ?? .longestSide
+            ) ?? .alternating
         splitRatioH =
             try container.decodeIfPresent(
                 Double.self,
--- a/Tests/KiwiDeskCoreTests/BspOverrideTests.swift
+++ b/Tests/KiwiDeskCoreTests/BspOverrideTests.swift
@@ -129,6 +129,11 @@ struct BspOverrideCommandTests {
     @Test("strategy override writes only that space")
     func strategyOverride() {
         let core = makeCore()
+        // Pin the global the untouched-space assertions read
+        // (#660), and pin it AWAY from the value the override
+        // writes: with both on the default the test cannot tell
+        // "only that space" from "every space" (#1181).
+        core.tiler.settings.bsp.strategy = .longestSide
         #expect(
             core.execute(
                 "bsp.set_strategy_override",
--- a/Tests/KiwiDeskCoreTests/LayoutBspTests.swift
+++ b/Tests/KiwiDeskCoreTests/LayoutBspTests.swift
@@ -65,9 +65,17 @@ struct BspLayoutTests {
 
     @Test("Third window splits the tall remainder vertically")
     func longestSideRecursion() throws {
+        // PINNED (#660/#1181), and this is the fixture that
+        // needs it most: on these bounds both strategies draw
+        // the same three frames, so the flip left the one test
+        // named for longest-side recursion silently running
+        // `.alternating` — and no test in the tree exercised
+        // `.longestSide` past depth 0 (code review, 2026-08-31).
+        var context = makeContext()
+        context.bsp.strategy = .longestSide
         let frames = layout.calculateGeometry(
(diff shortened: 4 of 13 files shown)


hajiboy95:
`guard-prover` report on the five repointed fixtures: **5 red-proofed, 0 inert, 0 vacuous.** Each reds on a sub-diff mutation of the thing it watches (dropping `strategy` from `encode`, from `init(from:)`, from `BspOverride.resolved(onto:)`; inverting the width/height comparison; and making `set_strategy_override` stamp the global).

Two results worth recording here:

- **The pins are load-bearing, and proven so.** With the `longestSideRecursion` pin removed, inverting the longest-side comparison goes **green** — that is exactly the regression the pin closes. Same for `strategyOverride`: without its global pinned away from the override's value, a command that stamps *every* space sails through, because "only that space" and "every space" are the same bytes when the two agree.
- **Coverage answered directly.** With a depth-scoped mutation (`depth == 0 ? (w >= h) : (w < h)`) run against the FULL suite: exactly two tests red — `longestSideRecursion` and `ratioIndependence`, both in this diff. With the pins removed, **zero**. So before this commit nothing in the tree exercised `.longestSide` past depth 0, and these two pins are now the only net over that branch.

Issue #1069, opened 2026-08-28
🐛 Reduce Motion is ignored by the .animation modifier (~47 sites, mostly the layout schematics)


hajiboy95:
### KiwiDesk version or commit

1.0.1 (00301725, `claude/roadmap-1-0-issue-1056-9caf77`)

### macOS version

macOS 26.6.2 (25G83)

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Turn on **System Settings ▸ Accessibility ▸ Reduce Motion**.
2. Open KiwiDesk Settings ▸ Layout Defaults.
3. Drag the window-count slider, or switch layout type.

The schematic previews spring and re-flow with full motion.
The same holds for the layer-strip and layer-chip hover fades,
the search field, the unsaved-changes footer and the onboarding
progress row.

### Expected vs. actual behavior

**Expected:** with Reduce Motion on, the app's own motion stands
down — the affordance stays, the travel goes. That is the rule
#989 established and `docs/ui-patterns.md` ▸ Interaction states
now states for users.

**Actual:** `withAnimation` honours it everywhere (#989, guarded
by `ReduceMotionGateTests`), but the **`.animation(_:value:)`
modifier** — the app's other way to start motion — is not
gated at ~47 sites and no guard watches them.

Measured 2026-08-28 by pointing `ReduceMotionGateTests`' needle
at `.animation(` as well: **49 sites surfaced**, of which two
were fixed by hand in #989 (`ProfilesSection`'s reorder, which
that issue names, and `InactiveDimmed`'s window fade). The
remainder breaks down as:

- **~40 in the layout schematics** — `GridSchematic`,
  `StackSchematic`, `TrackSchematic`, `ScrollingSchematic`,
  `BspSchematic`, `MonocleSchematic`, `FloatingSchematic`,
  `GapsDiagram` — all driving the shared
  `LayoutSchematic.damping`. One shared constant, so plausibly
  one fix.
- **~7 elsewhere**: `LayerStripEditor`, `ShortcutLayerChip`,
  `SettingsModeReveal`, `SettingsReveal`, `HoverCursor`,
  `SettingsFooter+Unsaved`, `SettingsSearchField`,
  `OnboardingProgressRow`, `WaitingDot`.

### Relevant output and logs

Not a runtime failure — a source property. To reproduce the
census, add `".animation"` beside `"withAnimation"` in
`ReduceMotionGateTests.entryPoints` and run
[...]

PR #1189, opened 2026-08-31
feat(gui): show the semantic version only in Settings

### LINKED ISSUE #1174: 💡 Show only the semantic version in the Settings GUI
### Use case

The Settings GUI shows the version as `1.1.1 (abc1234)` — the
commit hash is rendered on the General About card
(`Sources/KiwiDesk/Settings/Sections/GeneralSection+About.swift:12`)
and on the Home General card preview
(`Sources/KiwiDesk/Settings/HomeCardPreview.swift:131-142`).
Settings is used mostly by non-developers, and a hex blob after
the version is noise they cannot interpret.

### Proposed solution

The GUI renders the semantic version only — `v1.1.2` — on both
surfaces. The commit stays stamped and stays reachable for
support and debugging through `kiwidesk --version` and the
`version` command (`KiwiCore+Commands.swift:105-107` already
exposes `commit` as its own field).

Mechanically: `KiwiDeskVersion` (`App/KiwiDeskVersion.swift`)
grows a GUI-facing semantic-only string beside `displayString`,
and the two GUI call sites take it.

### Alternatives considered

Dropping the commit stamp entirely — rejected: the CLI and
support flows want it.

### Extra context

Ruled 2026-08-31 into 1.1.2 — patch-sized.

hajiboy95:
## Description

Settings showed the version as `1.1.1 (abc1234)`. Settings is
used mostly by non-developers, and a hex blob after the number is
noise they cannot interpret.

Both GUI surfaces — the General ▸ About card and the Home card
preview — now render the semantic version alone, through one
shared localized frame. The commit stays stamped and stays
reachable where support wants it: `kiwidesk --version` and the
`version` command, which already returned `commit` as its own
field.

## Related Issue(s)

Closes #1174

## Checklist

  (see AGENTS.md §3)
  required)

## How I verified

`swift build`, `swift test -q` (4306 tests), `scripts/lint.sh` —
green. Release build skipped locally (no concurrency or
`Sendable` change); CI's job runs it.

`VersionDisplayTests` holds both halves and I proved both red by
reverting the About site alone: `displayString` has exactly one
reader in the whole tree (`CLIMain.swift`), and exactly two files
carry the `general.version` frame.

## Notes for Reviewer

Two deliberate calls worth your eye:

**No new API.** The issue suggested `KiwiDeskVersion` grow "a
GUI-facing semantic-only string beside `displayString`" — but
`semantic` already *is* that string, so the call sites just take
it and `displayString` becomes the CLI's alone. Less surface, and
the guard is what keeps it that way.

**One key, shared.** About previously rendered the version with
no localized frame at all and no `v`; Home used `v%1$@`. They say
the same thing about the same value, so this is one key, not two
translations of one sentence — `home.card.general.version` was
renamed to `general.version` with `scripts/rename-key` (meaning
unchanged, so every translation carried over) and both surfaces
read it. About gains the `v` as a side effect, which is what the
issue asked for on both surfaces.

A `code-reviewer` pass is in flight; I will report anything it
turns up before this merges.


hajiboy95:
`code-reviewer` pass addressed. Its findings were all on my own guard, and two were real fail-opens I have now proven closed:

- **A comment could stand in for the call site.** Reverting the About site to `displayString` while leaving `// was: L("general.version", ...)` behind used to pass clause 2. `code(at:)` now strips `//` lines before matching — mutation re-run, both clauses red.
- **A wrapped member access was invisible.** `swift format` can split `KiwiDeskVersion` / `.displayString` across lines under the 79-char limit, which a per-line substring test cannot see (#1069's shape). The scan now strips whitespace too — I introduced that exact split at the Home site and confirmed clause 1 reds.

Also: the `displayString` docstring is cut to the contract plus its refs (§2.8 — the argument was reviewer-facing justification), the duplicated argument now lives in one place (the suite, where a reader arrives at it), and the suite states what its narrower `Settings/` root **trades**: a version rendered from Onboarding or the menu bar would not appear in clause 2 — clause 1 is what makes that safe, since it bars the commit from any such site whatever frame it used.

Confirmed by the reviewer and not by me, so recording it: `scripts/build-app.sh` reads `semantic` for both `CFBundleShortVersionString` and `CFBundleVersion`, so Sparkle's update prompt and the appcast were already semantic-only; and no doc or test asserts the old GUI shape (`docs/cli.md`'s `1.1.1 (abc1234)` documents `--version`, which is unchanged).

Issue #998, opened 2026-08-25
🐛 Navigating area→area states no keyboard focus destination


hajiboy95:
### KiwiDesk version or commit

`main` as of 2026-08-25 (40211da2).

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

Needs **System Settings ▸ Keyboard ▸ Keyboard navigation** on.

1. Open Settings and navigate from Home into an area. Focus lands on the back chip, as intended.
2. From there, follow a **cross-reference link into another area** (the `settingsNavigate` action — a link that jumps you from one area to a related one), or land in a different area from a search pick.
3. Press Tab.

### Expected vs. actual behavior

**Expected:** the new area states a focus destination, exactly as a Home→area push does. `.claude/rules/gui.md` ▸ the keyboard path requires it of **every** shape change: *"Every shape change states a focus destination, and the destination must be a control that is always DRAWN."*

**Actual:** nothing is stated. Focus was on a control in the previous area's subtree, that subtree is destroyed, and the assignment lands nowhere — so focus falls to the top of the window. That is the [#816](https://github.com/KiwiCanopy/KiwiDesk/issues/816) outcome reached by the second road `gui.md` names.

### Relevant output and logs

`Sources/KiwiDesk/Settings/SettingsHeaderBar.swift:112`:

```swift
.onChange(of: destination != nil) { _, pushed in
    if pushed { backChipFocused = true }
}
```

The trigger is keyed on the **Boolean** `destination != nil`, not on `destination`. Home→area flips it `false → true` and fires. Area→area leaves it `true` throughout, so `onChange` never fires and no destination is stated.

The correct spelling is already in the same file, **two modifiers below**, for an unrelated concern:

```swift
.onChange(of: destination) { _, _ in
    searchExpanded = false
}
```

So the fix is to key the focus statement on `destination` as well. Both spellings sitting seven lines apart is what let this pass review.

[...]

Issue #991, opened 2026-08-25
🐛 Navigating by mouse draws a keyboard focus ring


hajiboy95:
### KiwiDesk version or commit

`main` as of 2026-08-25 (1.0.0, 14ff44d and later).

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

Keyboard navigation does not need to be on, and VoiceOver is not involved.

1. Open Settings and click a card with the **mouse** to navigate into it.
2. Look at the back chip in the header.
3. Click the back chip to return, and look at the card you came from.

### Expected vs. actual behavior

**Expected:** navigating with the mouse draws no focus ring. macOS only shows a focus indicator when the user is navigating by keyboard — a mouse click does not move focus at all (text fields excepted, because they receive typing), so no ring appears.

**Actual:** the ring is drawn every time, because the destination is focused programmatically whether or not a keyboard was involved.

**This is not a request to hide the ring.** The focus indicator is a user-owned accessibility surface — Full Keyboard Access lets the user set its colour, contrast and size — and `docs/design-decisions.md` ▸ *a focus ring is the platform's* rules that an app must not answer those settings for the user. Suppressing it (`.focusEffectDisabled()`) would be exactly the cheap-direction fix that ruling warns about, and it would hurt the users who need the ring most.

The defect is one level up: **we move focus when the platform would not.** Fix that and the ring question disappears on its own, because there is nothing focused to ring.

### Relevant output and logs

The two shell-level sites, both unconditional:

- `Sources/KiwiDesk/Settings/SettingsHeaderBar.swift:113` — `if pushed { backChipFocused = true }`
- `Sources/KiwiDesk/Settings/HomeScreen.swift:147-149` — restores `nav.homeReturnFocus` on return

`.claude/rules/gui.md` ▸ the keyboard path requires this: *"Every shape change states a focus destination, and the destination must be a control that is always DRAWN"*, and names these two as the shell's own (push focuses the back chip, return restores `homeReturnFocus`). Sub-views state their own — `DeletionFocus.neighbour` (#816), `returningRow` in Profiles / Spaces / App Rules, `focusedChip` in `LayerStripEditor`.

[...]

Issue #816, opened 2026-08-11
🐛 Deleting a rule, profile, layer or palette drops keyboard focus to the top


hajiboy95:
### KiwiDesk version or commit

0.9.5 (`174a5c88`, branch `feat/awkward-states-keyboard`)

### macOS version

macOS 26.6.1 (25G76)

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

Filed from the dev machine during #678 Phase 4 pass 10. Any of
these four, by keyboard alone:

1. Open Settings ▸ Profiles and delete a saved profile.
2. Open Settings ▸ App Rules and delete a rule.
3. Open Settings ▸ Shortcuts and delete a layer from the layer
   strip.
4. Open Settings ▸ Advanced Colours and delete a saved palette.

Then press Tab and see where you are.

### Expected vs. actual behavior

Expected: focus moves to the next row, as turn 20a rule 4 states
("Deleting a space moves focus to the next row, never to the
top"), falling backward when the deleted row was last. Pass 10
implemented exactly this for the SPACES list —
`SpacesSection+Remove.swift` reads the neighbour before the
mutation and parks focus on that row's always-drawn layout
picker.

Actual: focus is dropped. The view that held it no longer exists,
nothing claims it, and the next Tab starts from the top of the
window — so a keyboard user re-walks the whole list after every
deletion.

### Relevant output and logs

Not applicable — no runtime failure. The four deletion sites,
each of which needs the shape `SpacesSection+Remove.swift`
already has:

- `Sources/KiwiDesk/Settings/Sections/ProfilesSection.swift`
  (delete profile) and `ProfilesSection+Broken.swift` (delete a
  broken one)
- `Sources/KiwiDesk/Settings/Sections/AppRuleRow.swift`
  (delete rule)
- `Sources/KiwiDesk/Settings/Sections/LayerStripEditor.swift`
  (delete layer)
- `Sources/KiwiDesk/Settings/Components/Colors/PaletteShelf.swift`
  (delete palette, currently reachable only from the context
  menu / its mirrored accessibility actions)

Two things pass 10 learned that this work should not re-learn:

1. **The destination must be a control that is always DRAWN.**
   The first cut bound the spaces list's return to
   `offeredCustomizeButton`, which `SpaceOverrideOffer.isOffered`
   only draws in Power User mode or when the profile already
   carries an override. An absent mode preference means Simple,
   so on a fresh install the assignment named a value no view
[...]

PR #1192, opened 2026-08-31
fix(gui): land arrival focus on the page, not the back chip

### LINKED ISSUE #996: 🐛 Arriving at a Settings page lands keyboard focus four Tab stops from the page

### macOS version

macOS 26.6.2

### Mac model type

### Steps to reproduce

Needs **System Settings ▸ Keyboard ▸ Keyboard navigation** on, per the first row of `docs/accepted-limitations.md`.

1. Open Settings, navigate into any area (Spaces, Colours & Motion, …).
2. Press Tab repeatedly and count how many presses it takes to reach the first control **of the page**.

**Expected:** arriving at a page puts the keyboard within a press of that page's content.

**Actual:** roughly **four presses**. Arrival focuses the header's back chip (`SettingsHeaderBar.swift:112` — `.onChange(of: destination != nil) { if pushed { backChipFocused = true } }`), and the rest of the header sits between it and the content: the search field, the profile chip, the help button and the Simple/Power segmented control. A keyboard user pays that on **every** navigation.

Owner-observed 2026-08-25.

### Relevant output and logs

**Why the back chip was chosen, and why the obvious fix is not free.** `.claude/rules/gui.md` ▸ the keyboard path requires that "Every shape change states a focus destination, and the destination must be a control that is always DRAWN". The back chip is the one control guaranteed to exist on every page, which is exactly why it was picked. [#816](https://github.com/KiwiCanopy/KiwiDesk/issues/816) shipped the failure of ignoring that clause: focus was bound to a mode-gated affordance, went to a value no view claimed, and fell to the top of the window.

So "focus the page's first control" is **not** a safe fix as stated — it would mean ~10 destinations each naming an always-drawn target, and each is a chance to repeat #816.

**A cheaper shape that may already exist.** The detail pane is `contentColumn` in `Sources/KiwiDesk/Settings/SettingsView+Detail.swift`, which renders `detail(model.destination)`. Every section's root is already its own `ScrollView`, and it is already accessibility-labelled with the area's title — the code comments that VoiceOver otherwise lands on it as a bare "scroll area" ([#812](https://github.com/KiwiCanopy/KiwiDesk/issues/812) session 2). That container is always drawn by construction, exists once per destination rather than ten times by hand, and Tab from it would land on the page's first real control.

Whether that container can hold focus, and whether landing there announces something sensible under VoiceOver, is the design question this issue opens. `ui-designer` should rule the shape before it is built.

**Two things this must not break:**

- The **search** path. A search hit navigates and reveals a specific control; that flow states its own destination and must keep it.
- The **return** path. `HomeScreen` restores `nav.homeReturnFocus` so going back lands on the card you came from — correct, and untouched by this.

- [x] None that I know of

### Extra context

Found 2026-08-25 while scoping [#991](https://github.com/KiwiCanopy/KiwiDesk/issues/991) (mouse navigation drawing a keyboard focus ring). The two are **independent** and should stay separate lanes:

- #991 is about *whether* focus is set at all when the navigation came from a click — the input-source gate is correct whatever the destination turns out to be.
- This issue is about *which* control the destination should be when the navigation did come from the keyboard.

They do share a device session, so it is worth ruling this one before implementing #991 — otherwise the same keyboard-navigation check gets run twice.

**This needs device verification with keyboard navigation ON.** Nothing headless separates "a destination was named" from "focus arrived" — `gui.md` says so, and a source-scanning needle can only prove the former.

hajiboy95:
## What & why

fixes #996

Arriving at a Settings area focused the header's back chip, leaving five header stops between a keyboard user and the page they asked for — **six presses to the first field**, paid on every navigation (owner measurement on the issue).

The destination moves to the **content pane**. It is drawn by construction of the branch it lives in, exists once rather than twelve times by hand, and already carried the area's accessibility label — so the focus target and the named target are one element and cannot drift.

The shape was ruled by `ui-designer` before it was built, as the issue asked. Focusing each section's first control instead would have meant twelve destinations each owing a judgement that its target is drawn in both modes, at all three widths and under `editingStoredProfile` — twelve chances to repeat #816, which shipped once already.

**The one thing worth reviewing closely:** the raise lives on `structuredShell`, *not* on the pane it focuses. `onChange` does not fire on a view's first appearance, and the pane is created by the Home→area transition — so the same closure hung on the pane would state nothing on exactly the navigation #998 was about. It keeps #998's value-keying for the other half of that lesson.

The label is hoisted out of `model.destination?.title ?? ""`; an empty fallback was tolerable on a label nobody could focus, but a nameless Tab stop is worse than the presses it replaces.

The back chip keeps `⌘[` and Escape (both already documented) and stays Shift-Tab reachable.

## Verification

Full gate green: `swift build`, `swift test -q` (4315 tests / 819 suites), `scripts/lint.sh`, plus the site build because `docs/` is one of its inputs. Release build skipped — no concurrency or `Sendable` surface; read CI's `Release Build` job.

`guard-prover` (isolated worktree) red-proofed all four sub-diff mutations: `.focusable()` removed with `.focused()` left in place, `.focused()` removed with `.focusable()` left, the raise re-keyed on `destination != nil`, and the raise moved out of the shell file.

It also found a **real gap**, reported rather than papered over: moving the raise *within* `SettingsView.swift` onto the `selection == nil` branch — firing from the subtree the transition is tearing down — kept the guard green. That is closed on the stacked #991 branch, where the needle grows to carry its neighbour and pin the raise to `structuredShell`'s own chain.

## Not verified here

**Focus ARRIVING is not something a needle can show.** Both shell needles say only that a destination was named; `gui.md` is explicit that nothing headless separates "drawn" from "focusable". The device confirm with **keyboard navigation ON** is outstanding — the checks to run are listed on the issue, and the discriminator is: from arrival, one Tab must reach the section's first control. If the assignment fell through, that first Tab lands on the profile chip instead.

Opened as a **draft** deliberately: 1.1.2 is being tagged in a parallel session, and nothing may reach `main` before that tag.

### DIFF
--- a/.claude/rules/gui.md
+++ b/.claude/rules/gui.md
@@ -875,14 +875,24 @@ claim; the obligations a change here takes on:
   behaviour the rule exists to prevent, and invisible to a
   source-needle guard, which is why the wirings in
   `KeyboardActionParityTests` are keyed on their use sites.
-  The shell states two of these itself (push focuses the back
-  chip, return restores `nav.homeReturnFocus`); a sub-view the
-  shell cannot see states its own, and a deletion reads its
+  The shell states two of these itself (push focuses the
+  CONTENT PANE, return restores `nav.homeReturnFocus`); a
+  sub-view the shell cannot see states its own, and a deletion
+  reads its
   neighbour through the one `DeletionFocus.neighbour` rule
   (#816, PR #842) — BEFORE the mutation, or it names whichever
   row slid into the gap. Naming a destination is a different
   claim from focus arriving there, and on a machine without
   keyboard navigation the two outcomes look identical.
+- **A shell statement hangs on the view that OUTLIVES the
+  transition, never on the one the transition builds** (#996).
+  `onChange` does not fire on a view's first appearance, so a
+  raise hung on the pane it focuses states nothing on the one
+  navigation that creates that pane — #998's defect at the
+  opposite polarity. That is why the arrival raise sits on
+  `structuredShell` rather than beside the `.focused` it drives,
+  which reads like the tidier placement and is the broken one.
+  `KeyboardActionParityTests+Shell` pins the raise to that chain.
 
 ## Colour (#678 turn 16b)
 
--- a/Sources/KiwiDesk/Settings/Components/Common/ClickBornFocus.swift
+++ b/Sources/KiwiDesk/Settings/Components/Common/ClickBornFocus.swift
@@ -28,7 +28,10 @@ enum ClickBornFocus {
     /// focus change is observed.
     @MainActor static var isClickBorn: Bool {
         if NSEvent.pressedMouseButtons != 0 { return true }
-        switch NSApp.currentEvent?.type {
+        // `NSApp?`, never `NSApp`: it is implicitly unwrapped
+        // and nil in a process with no `NSApplication`, where
+        // the force-unwrap traps (crash, 2026-09-01).
+        switch NSApp?.currentEvent?.type {
         case .leftMouseDown, .leftMouseUp, .rightMouseDown,
             .rightMouseUp, .otherMouseDown, .otherMouseUp:
             return true
--- a/Sources/KiwiDesk/Settings/Components/Common/ClickBornFocus.swift
+++ b/Sources/KiwiDesk/Settings/Components/Common/ClickBornFocus.swift
@@ -1,0 +1,39 @@
+import AppKit
+
+/// Whether the focus change now landing was caused by the mouse.
+///
+/// macOS 26 gives a `.focusable()` custom view keyboard focus on
+/// a click, where the platform's own controls take none — so a
+/// ring appears for a user who never asked for one (#991's
+/// defect; owner eye-confirm on the Settings pane, 2026-09-01).
+/// Refuse the focus and there is nothing to ring, which is the
+/// direction `docs/design-decisions.md` ▸ *a focus ring is the
+/// platform's* requires.
+///
+/// The PREDICATE is shared; the `onChange` that consults it is
+/// spelled at each site on purpose. Folded into a `ViewModifier`
+/// taking the `FocusState` binding it silently never fires —
+/// reading `wrappedValue` there establishes no dependency, so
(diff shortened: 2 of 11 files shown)


hajiboy95:
## Review round complete — one finding lands on this branch

`code-reviewer` and `architect-reviewer` both flagged the same thing against this change, and it is fixed here rather than on the stacked PR so that **neither branch can land a false rule alone**:

**`.claude/rules/gui.md` still said the shell's push focuses the back chip.** That file auto-loads whenever anyone edits this tree, so the next author would have started from a premise this PR falsifies and re-implemented the back-chip raise. The bullet now names the content pane, and carries the obligation this change discovered:

> A shell statement hangs on the view that **outlives** the transition, never on the one the transition builds — `onChange` does not fire on a view's first appearance, so a raise hung on the pane it focuses states nothing on the one navigation that creates that pane.

That is the trap this PR's own design note is about, written where it will be read.

`architect-reviewer` on the shape, quoted because it is the answer to "why not just focus each section's first control": the twelve-destination alternative *"is not a shape, it is a debt schedule"* — a census can count twelve modifiers, but no census can tell you the twelfth is attached to a row Simple mode hides.

Two things it confirmed rather than found:

- **The `if let destination` hoist is clean for the reveal path.** `contentColumn` is reached only through `detailPane`, which the shell builds only when `selection != nil`, so the new binding cannot be false while the subtree is mounted; `.onAppear`, `pendingScroll`, `revealTarget` and the flash all fire exactly as before.
- **Core boundaries are not crossed.** Nothing outside `Sources/KiwiDesk`, `Tests/KiwiDeskGuiTests` and `docs/`.

### One thing added to the device checklist as a result

`code-reviewer` raised a question it could not settle headlessly, and it is now on the issue: on an **area→area** navigation the focused view is torn down and rebuilt in the same update the raise assigns in, so whether the assignment survives the teardown is ordering-dependent. **Confirm area→area specifically** (follow a cross-reference row), not only Home→area — that is #998's own path.

The `guard-prover` gap this PR's body mentioned is closed on the stacked branch, by reading the raise's depth rather than its neighbours.

Issue #997, opened 2026-08-25
🐛 A segmented control costs one Tab stop per segment instead of one for the control


hajiboy95:
### KiwiDesk version or commit

`main` as of 2026-08-25 (40211da2).

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

Needs **System Settings ▸ Keyboard ▸ Keyboard navigation** on.

1. Open Settings on any page holding a segmented control — the header's Simple / Power User toggle is the easiest.
2. Tab through the page and count the stops the control costs.

### Expected vs. actual behavior

**Expected:** one Tab stop for the whole control, with **arrow keys** moving between segments. That is what a native segmented control does, and what a `Picker` in `.segmented` style gives free.

**Actual:** **one Tab stop per segment.** The header's two-segment toggle costs two stops; a three-segment control costs three.

Owner-observed 2026-08-25 while counting the stops in [#996](https://github.com/KiwiCanopy/KiwiDesk/issues/996): arriving at a page and reaching its first field takes six presses — search field, profile chip, help button, **Simple, Pro**, first field — of which two should have been one.

### Relevant output and logs

`Sources/KiwiDesk/Settings/Components/Common/SegmentedPicker.swift` builds the track as a `ForEach(options.indices)` where each segment is its own `Button`. Every `Button` is a focus stop, so the stop count is the segment count.

**The VoiceOver half was already re-earned; the keyboard half was not.** The same file puts `.accessibilityElement(children: .contain)` on the track and gives it the group's choice as its value, with a comment saying it should announce "Position, Top" *"the way a native segmented control says it"*. So the intent was already to behave like the native control — this is the other half of that same intent, and `.claude/rules/gui.md` names the obligation directly: *an AppKit control re-earns the focus, keyboard, VoiceOver and `isEnabled` its SwiftUI twin gave free* (the `LinkedCaptionHitTests` rule).

**Blast radius: 30 call sites**, not just the header — `GeneralSection`, `BehaviorSection`, the App Bar and Space Bar cards, `SlotSizeRows`, the Layout cards and more. Every one of them costs its segment count in Tab stops on every keyboard traversal of its page.

### Expected vs. actual — what the fix has to preserve

- **Clicking a segment directly** must keep working; it is the primary interaction.
- **The existing VoiceOver announcement** is already correct and must not regress — the track announces the group and its current choice, and each segment carries its own traits.
[...]

PR #1193, opened 2026-08-31
fix(gui): state a focus destination only when the keyboard asked

### LINKED ISSUE #991: 🐛 Navigating by mouse draws a keyboard focus ring

### LINKED ISSUE #996: 🐛 Arriving at a Settings page lands keyboard focus four Tab stops from the page

hajiboy95:
## What & why

fixes #991

> **Stacked on [#1192](https://github.com/KiwiCanopy/KiwiDesk/pull/1192) (#996).** Base is `fix/settings-arrival-focus`, so the diff here is only #991's. It retargets to `main` on its own when #1192 lands.

Navigating by mouse drew a keyboard focus ring, because the shell's two focus statements fired whether or not a keyboard was involved.

**This is not fixed at the ring, and that is the whole point of the issue.** A focus ring is the user's own accessibility surface — Full Keyboard Access owns its colour, contrast and size — so `.focusEffectDisabled()` is the cheap-direction fix `docs/design-decisions.md` ▸ *a focus ring is the platform's* forbids, and it would hurt the users who need the ring most. The defect is one level up: we moved focus where macOS would not have. Move focus only when the platform would, and there is nothing focused to ring.

## The seam

The issue's own condition: *"one seam consulted by every focus-destination site, not a per-site condition, or the next site will forget it."*

The input source is recorded in **one place — `destination`'s `didSet`** — rather than at each navigation verb. Every path already passes through that property (the cards, the cross-reference links, a search hit, the back chip, Escape, and the programmatic repairs), so a path added later cannot forget it. Reading it there also reads it while macOS is still dispatching the event that caused the navigation.

`.keyDown` is the whole keyboard set that reaches a navigation — Space and Return on a focused control, Return on a search result, Escape and `⌘[` all arrive as one. A click is a mouse event and a deep link is no event at all; neither moves focus, and Tab then restarts from the top of the window, which is what macOS does and what the owner confirmed in System Settings.

On the Home restore, **the clear stays outside the gate**: a mouse pop must not restore focus, but it must still empty the slot, or a stale card leaks into the next unrelated appear.

## Also closes #996's guard gap

`guard-prover` on #996 found that naming the file alone let the raise be moved onto the `selection == nil` branch — firing from the subtree the Home→area transition is tearing down — with the guard still green. The needle now carries its neighbour (`.onExitCommand`), pinning it to `structuredShell`'s own chain.

## Verification

Full gate green: build, `swift test -q` (4317 tests / 820 suites), lint, site build. Release build skipped — no concurrency surface.

New `SettingsInputSourceSeamTests` holds both halves of the one-seam claim: only `SettingsInputSource` reads `NSApp.currentEvent` under the Settings tree (reach stated in the test — `StatusItemController+Menu.swift` reads it outside that tree and is deliberately out of scope), and both focus statements consult the recorded value. A seam nothing consults is not a seam, which is why the second test exists.

`guard-prover` is running on this branch; findings will be posted here and addressed before this leaves draft.

## Not verified here

A scan can say the gate is spelled at both statements; **it cannot say a ring stopped being drawn.** Device confirm outstanding, batched with #996 and #997 into one keyboard-navigation sitting. The check: click a card with the mouse — no ring anywhere; then Tab — it starts from the top of the window, not from the page.


hajiboy95:
## Review round complete — 1 blocker, 4 major, 2 minor, all addressed

`code-reviewer` + `architect-reviewer` in parallel, then `guard-prover` on the changed assertions. What changed as a result:

### Blocker — the predicate was backwards (fixed, `ae40797b`)

`NSApp.currentEvent?.type == .keyDown` is a **positive** test, so everything that is neither a click nor a literal key event fell to "not keyboard" and stated no destination at all. Two paths land there, and the first is the population the rule exists for:

- **VoiceOver presses a control through `AXPress`, with no `NSEvent` behind it.** Its cursor follows keyboard focus — measured, and recorded in `docs/accepted-limitations.md` — so a VoiceOver user navigating from a Home card was left with the cursor on a card the transition had just destroyed. That is the top-of-window outcome gui.md's keyboard path exists to prevent, reached by a new road. My change would have introduced it.
- **`NSApp.currentEvent` is the last event *retrieved*, not one cleared after dispatch.** A programmatic write following a key press read that stale key event and claimed the keyboard had asked. My docstring asserted the opposite; it was simply wrong.

Now it **refuses a positive mouse event and nothing else**, which fixes both and fails *toward* stating a destination — the safe direction, since a click is the only thing this issue measured. Renamed `movesFocus` / `navigationMovesFocus`, because the old names claimed more than the test does.

### Major — `gui.md` still taught the old rule (fixed)

The owning rule file said the shell's push focuses the back chip, which #996 falsifies, and auto-loads for the next author touching this tree. #996's half of the correction is on **its own branch** (so neither PR can land a false rule alone); this branch adds the new obligation with its guard, refreshes the `AGENTS.md` §5 row, and fixes `docs/accepted-limitations.md`'s VoiceOver row, which asserted the same thing.

### Major — §2.8 comment budget (fixed)

Twelve comment lines on one `onChange`, a fifteen-line file-level argument naming a rejected alternative, a ten-line docstring of design history. All over budget and, for the rejected alternative, explicitly the kind of thing §2.8 sends to the rule file. Cut to the constraint plus the ref; the argument now lives in gui.md, where the reviewer wanted it anyway.

### Minor — the latch outlived its visit (fixed)

`resetSurfaces()` now clears it. It is only meaningful to a statement fired by the write that set it, so a window re-shown must not inherit the last visit's verdict.

### `guard-prover`: 5 mutations red-proofed, 1 residual hole found and closed

Every gate-drop, the clear moved inside the gate, the deleted `didSet`, and the raise moved out of `structuredShell` all red with the right message.

It then found that moving the raise **together with its neighbour** onto the `selection == nil` branch stayed green — the same defect one step deeper. Closed in `a67b3d95` by reading **depth** instead of order: the raise must sit outside `chrome(width) { … }`, which outlives both branches. I replayed the prover's exact mutation against the new guard to confirm it reds.

### Consciously dismissed

- **Fold #997's `pressedMouseButtons` refusal onto this seam.** `code-reviewer` wanted it here; `architect-reviewer` argued against merging, since "a mouse button is physically down" and "the current event is a key press" are different questions and the second refuses programmatic focus, which the first must allow. Both are satisfied by leaving them separate and *counting* — #1194 adds an exact-set census, so the third copy is a decision rather than an omission, and its failure message sends the next author here. Coupling two otherwise-independent PRs to share five lines was the worse trade.

### Still outstanding

The device confirm. Note the reviewer's specific ask, which I have added to the issue: **confirm area→area** (a cross-reference link), not only Home→area — the focused view is torn down and rebuilt in the same update the raise assigns in, and whether the assignment survives is ordering-dependent and headlessly unprovable.

Issue #812, opened 2026-08-10
💡 Deep VoiceOver hardening for the redesigned Settings (pre-1.0)


hajiboy95:
### Use case

Using the redesigned Settings window with VoiceOver as the
primary interface. #678 Phase 4 pass 10 ships the keyboard half
(full keyboard reachability) plus the baseline AX the census
guards already hold — labels on every census-named row, combined
row elements, no context living in chrome VoiceOver cannot see.
What it deliberately does NOT ship is the deep choreography:
rotor-friendly grouping and heading structure per area, announced
value changes on live controls (sliders, the draft diff),
sensible reading order through the two-column detail shell and
its floating save pill, the keyboard preview board narrating key
states, and a spoken equivalent for every schematic caption.

### Proposed solution

A dedicated VoiceOver hardening pass over the frozen Settings
surfaces, run BEFORE 1.0 and AFTER the redesign's layout passes
stop moving them (owner ruling 2026-08-10: choreography polished
before the surfaces freeze is work done twice). Scope: a full
VoiceOver walkthrough of every destination in both Settings
modes, fixing grouping, ordering, announcements and rotor
navigation; extend the existing AX guards where a fix class
recurs. The `search.result_mode_ax` lesson is the bar: a frame
must carry its own context, because on-screen chrome (capsules,
icons, position) is invisible to the reader.

### Alternatives considered

- Folding it into pass 10 now: rejected — passes 8/9/11 still
  move the very surfaces it would polish, and the beta's single
  user does not use VoiceOver, so the work would churn twice for
  no current reader.
- Skipping it entirely: rejected — gui.md's north star names
  VoiceOver a non-negotiable system convention, and an App Store
  1.0 will be judged on it.

### Extra context

Deferred from #678 Phase 4 pass 10 by owner ruling 2026-08-10
(pass 10 = keyboard + baseline AX only; sequencing lives in
plan/settings-redesign/implementation-plan.md). Dev machine:
macOS 26 (Darwin 25.6.0), no other AX tools running beside
KiwiDesk itself.

hajiboy95:
Lane record for the cold start (branch `feat/voiceover-hardening`, 14 commits, PR to follow). Device-verified with the owner across three VoiceOver sessions, 2026-08-24.

## What shipped

[...]

Issue #754, opened 2026-08-06
💡 Gaps & Borders: one Borders card, linked width and corner, Focus first


hajiboy95:
### Use case

Gaps & Borders draws three strokes — the focus ring, the drag
ghost and the drop zone — and gives each its own width, its own
corner and its own inside/outside alignment. In practice a user
picks one look and wants all three to match it: a 3 pt ring and a
1 pt ghost is not a preference anyone holds, it is what happens
when three sliders are set independently and two are forgotten.

So the page asks the user to make the same decision three times,
and the ordering makes it worse: `DragVisualsEditor` is stacked
above `FocusBorderEditor`, i.e. the two strokes seen during a
drag lead the one seen all day.

Current rows (`GapsBordersRowOrder`):

- `.focusBorder` — `borderWidth`, `borderCorner`, `borderGlow`,
  `borderGlowSizeAuto`, `borderGlowSize`,
  `borderFitGapsExtraSpacing`, `borderFitGaps`
- `.dragAndDrop` — `dragCornerRadius` (already shared),
  `dragGhostBorderWidth`, `dragGhostBorderAlignment`,
  `dragDropZoneBorderWidth`, `dragDropZoneBorderAlignment`

### Proposed solution

**1. A Borders header card on top, holding the shared decisions.**
Above all three sections:

- **Border width** — one slider driving the focus ring, the ghost
  and the drop zone.
- **Corner radius** — same, one master. `dragCornerRadius` is
  already shared between ghost and drop zone, so this only folds
  `borderCorner` in beside it.
Border alignment does **not** join them — see 5 below; it leaves
the GUI entirely.

Vocabulary: it is **width**, not thickness. A stroke has a width
(`border.set_width`, `drag.*_border_width`); a bar has a
thickness. `config-vocabulary.md` owns that pair and the label
has to follow it.

**2. A "Use one width for all borders" toggle, default on.** With
it on, the per-section width/corner sliders stay **visible and
greyed**, showing the master's value — grey, don't hide (#171).
The grey is the hint that the master owns them, so no extra
caption is needed. Turn it off and the three sections' own
sliders go live again, master included as the focus ring's own.

**3. Focus border first.** Order becomes: Borders header card →
Focus border → Drag ghost → Drop zone → Sticky mark. The most
frequently seen stroke leads, and it is the one whose look the
other two are matching.

**4. No new "more" accordion — use the tier that exists.**
[...]

PR #1194, opened 2026-08-31
fix(gui): a segmented control costs one Tab stop, not one per segment

### LINKED ISSUE #997: 🐛 A segmented control costs one Tab stop per segment instead of one for the control

hajiboy95:
## What & why

fixes #997

`SegmentedPicker` built its track as a `Button` per segment, and every `Button` is a focus stop — so the header's two-segment mode switch cost two Tab presses and a three-segment control cost three, **across 30 call sites**. A native segmented control costs one stop for the whole thing and moves the choice on ← / →.

The track takes the stop the segments give up: `.focusable(isEnabled)` on the track, `.focusable(false)` on each segment.

**Both halves are load-bearing.** With only the first, the control *gains* a stop instead of trading N for one — the reported defect plus one press per control. The guard needles both, and the negative clause beside them covers the same escape from the other side (SwiftUI takes the last `.focusable` on a chain).

The segments stay `Button`s, so everything the issue said must not regress is untouched by construction: the click path, the per-segment `.isSelected` trait, the group's #812 announcement, and `isEnabled`.

## Two judgement calls worth a look

**← / → move the selection, not a highlight.** That is what the native control does, and a highlight that moved without selecting would announce a choice the binding never took. ↑ / ↓ are deliberately *not* consumed, so vertical traversal still carries you out of the row. Arrows clamp rather than wrap; an unmatched selection (nil, #754) lands on the first option rather than stepping from an index it does not have.

**The track refuses click-born focus**, the way `SettingsSlider` already does. Without it, making the track focusable would draw a keyboard focus ring on every mouse click on a segment — which is #991's defect, introduced here by this fix. It is in scope because this change would otherwise cause it.

## Verification

Full gate green: build, `swift test -q` (4319 tests / 820 suites), lint, site build (`docs/` touched). Release build skipped — no concurrency surface.

`guard-prover` (isolated worktree) red-proofed 8 sub-diff mutations across the 4 new tests — wrap-instead-of-clamp, the nil case treated as 0, the removed empty guard, `.focusable(false)` deleted *and* flipped to `true`, the track's focusable deleted, an arrow wiring deleted, and the shared arithmetic replaced by a behaviourally identical inline copy. It also confirmed `swift format` reflowing a needled call across lines does not red the needles.

**It found a real fail-open**, now fixed: read file-wide, every needle stayed green while the focus pair was *moved* onto the segment — this defect restored under a green guard. Each needle is now read inside the declaration that must carry it (`SourceScan.declarationBody`), and I reproduced the prover's exact mutation against the fixed guard to confirm it reds.

## Not verified here

A scan can say the control builds one focusable element instead of N; **it cannot say focus behaves.** Device confirm with keyboard navigation ON is outstanding, batched with #996/#991 into one sitting. The checks: the header's Simple | Power User switch costs **one** Tab press, ← / → move between segments while focused, clicking a segment draws no ring, and a disabled picker is skipped entirely.

### DIFF
--- a/Sources/KiwiDesk/Settings/Components/Common/SegmentedPicker.swift
+++ b/Sources/KiwiDesk/Settings/Components/Common/SegmentedPicker.swift
@@ -1,3 +1,4 @@
+import AppKit
 import SwiftUI
 
 /// Capsule segmented picker with animated sliding accent pill (#68).
@@ -9,6 +10,7 @@ struct SegmentedPicker<Value: Hashable>: View {
     private let help: String?
     @Namespace private var pillSpace
     @State private var hoveredIndex: Int?
+    @FocusState private var focused: Bool
     @Environment(\.accessibilityReduceMotion)
     private var reduceMotion
     @Environment(\.isEnabled) private var isEnabled
@@ -75,6 +77,27 @@ struct SegmentedPicker<Value: Hashable>: View {
                 lineWidth: 0.5
             )
         )
+        // ONE Tab stop for the whole control, the way a native
+        // segmented control behaves — the segments opt out below,
+        // and the track takes the stop they gave up (#997).
+        .focusable(isEnabled)
+        .focused($focused)
+        .onChange(of: focused) { _, now in
+            // Refuse click-born focus on macOS 26, the same
+            // refusal `SettingsSlider` carries: without it every
+            // click on a segment draws a keyboard focus ring
+            // (#991's family, which this stop would otherwise
+            // introduce here).
+            guard now, NSEvent.pressedMouseButtons != 0 else {
+                return
+            }
+            focused = false
+        }
+        // ← / → only. ↑ / ↓ are left alone so vertical traversal
+        // still leaves the row, which is what the native control
+        // does too.
+        .onKeyPress(.leftArrow) { move(-1) }
+        .onKeyPress(.rightArrow) { move(1) }
         .accessibilityElement(children: .contain)
         // Group value announcement for accessibility (#812, 2026-08-24).
         .accessibilityValue(
@@ -83,6 +106,24 @@ struct SegmentedPicker<Value: Hashable>: View {
         .onChange(of: isEnabled) { _, now in
             if !now { hoveredIndex = nil }
         }
+    }
+
+    /// Moves the SELECTION one segment, which is what a native
+    /// segmented control does on an arrow — a highlight that
+    /// moved without selecting would announce a choice the
+    /// binding never took.
+    private func move(_ direction: Int) -> KeyPress.Result {
+        guard isEnabled,
+            let next = SegmentedPickerKeys.step(
+                from: selectedIndex,
+                by: direction,
+                count: options.count
+            )
+        else { return .ignored }
+        // Clamped, so the end segments swallow their own arrow
+        // rather than letting it walk out of the control.
+        if next != selectedIndex { select(options[next].value) }
+        return .handled
     }
 
     private func segment(
@@ -100,6 +141,11 @@ struct SegmentedPicker<Value: Hashable>: View {
             )
         }
         .buttonStyle(.plain)
+        // The stop this Button gave free is the defect: N
+        // segments cost N Tab stops where the native control
+        // costs one. Click, VoiceOver and `isEnabled` all stay —
+        // only the focus stop goes, to the track (#997).
+        .focusable(false)
         .onHover { inside in
             updateHover(
                 inside: inside,
--- a/Sources/KiwiDesk/Settings/Components/Common/SegmentedPicker.swift
+++ b/Sources/KiwiDesk/Settings/Components/Common/SegmentedPicker.swift
@@ -1,4 +1,3 @@
-import AppKit
 import SwiftUI
 
 /// Capsule segmented picker with animated sliding accent pill (#68).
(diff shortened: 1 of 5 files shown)


hajiboy95:
## Review round complete — 2 major, 1 minor, all addressed

### Major — the click-born refusal was an uncounted second copy (fixed)

This change makes the **second** verbatim copy of the `NSEvent.pressedMouseButtons` refusal (`SettingsSlider` has the first), and nothing was counting — so the third would have been an omission rather than a decision.

The two reviewers split on the remedy, and the split is the interesting part. `code-reviewer` wanted it folded onto #991's input-source seam. `architect-reviewer` argued specifically against merging: *"a mouse button is physically down" and "the current event is a key press" are different questions, and merging them would wrongly refuse programmatic focus.*

Both are satisfied by leaving the code separate and **counting** it. The new census holds the reader set exactly — a focusable control that skips the refusal reds, and one that adds it without needing it reds too — and its failure message tells the next author to home the two together at the third site rather than adding an entry. Coupling this PR to #1193 to share five lines was the worse trade.

### Major — dismissed with evidence: the track is named everywhere

`architect-reviewer` flagged that the track now carries `.accessibilityValue` with no `.accessibilityLabel`, making the control's one focus stop a nameless one. Checked rather than assumed: `labeledTrack` applies `.accessibilityLabel(label)` whenever the picker is constructed with a label, and both call sites that omit the label — `SettingsHeaderBar`'s mode segment and `IconPicker` — carry their own `.accessibilityLabel`. Zero unnamed sites today, so there is nothing to fix. Worth confirming by ear in the sitting all the same, since only a device says what it announces.

### Minor — the inert guard is gone (fixed)

`move(_:)`'s `guard isEnabled` could not fail: the track is `.focusable(isEnabled)`, so a disabled control receives no key press to answer, and the `.ignored` return was unreachable. Removed, with what makes it unreachable stated at the site — and the needle on `.focusable(isEnabled)` is what keeps that true.

Also trimmed three comment blocks to the §2.8 budget.

### `guard-prover`: 8 sub-diffs red-proofed, 1 fail-open found and closed

Wrap-instead-of-clamp, nil-treated-as-0, the removed empty guard, `.focusable(false)` deleted *and* flipped to `true`, the track's focusable deleted, an arrow wiring deleted, and the arithmetic replaced by an identical inline copy — all red. Reflowing a needled call across lines does **not** red them, which was the specific tax I asked it to check for.

The fail-open: read file-wide, every needle stayed green while the focus pair was **moved onto the segment** — SwiftUI takes the last `.focusable`, so the segments become stops again and the track does not, which is this issue's defect restored under a green guard. Each needle is now read inside the declaration that must carry it, plus a negative clause covering the same escape from the other side. I replayed the prover's exact mutation against the fixed guard: it reds with two issues.

### Question carried to the sitting

`move(_:)` returns `.handled` even when clamped at an end. `code-reviewer` judged this right — `NSSegmentedControl` consumes ← / → as key view, and returning `.ignored` at the ends would make the first left-arrow at segment 0 behave differently from the second. The cost is that a horizontally scrollable ancestor cannot be arrow-driven while the picker holds focus, which also matches the native control. Flagging it because it is a taste call, not a correctness one.

PR #1196, opened 2026-08-31
feat(gui): picking an app adds its rule immediately

### LINKED ISSUE #1172: 💡 App Rules: picking an app adds the rule immediately
### Use case

Adding an app rule today is a two-step: pick the app in the
selector, then click the separate "Add app rule" button
(`Sources/KiwiDesk/Settings/Sections/AppRulesSection.swift:126-151`).
The button is pure friction — picking an app already says
everything. The intermediate picked-but-not-added state also
renders without the app icon (the icon shows in the dropdown and
again after adding, but not in between), which reads as a glitch.

### Proposed solution

Selecting an app from the installed-apps dropdown adds the rule
immediately; the "Add app rule" button goes away. The
intermediate state disappears, and the icon inconsistency
disappears with it — one change, both complaints gone.

Exception: the picker's "Custom…" bundle-id text field
(`AppRuleControls.swift`) cannot add-on-select — free text keeps
a commit action (Return, or a small inline confirm) for that
path only.

### Alternatives considered

Fixing only the missing icon in the intermediate state —
pointless once the state itself is removed.

### Extra context

Ruled 2026-08-31 into 1.2.0. Row/census obligations per
gui.md apply to whatever the add row becomes.

hajiboy95:
## What & why

`fixes #1172`

Adding an app rule was a two-step: pick the app, then click a separate **Add app rule** button. The button was pure friction — picking an app already says everything — and the intermediate picked-but-not-added state was also the one state that drew no icon, which read as a glitch.

Both complaints go with one change. The picker commits what it picked, the button is gone, and the state between them no longer exists to render wrongly.

## The exception, and why it is one

Free text keeps a commit, as the issue rules. The reason is worth stating: **every keystroke of `com.apple.Safari` is a prefix of `com.apple.Safari`**, so the typed path has no moment that means "this is the app".

It keeps a commit on **both channels** — Return *and* a visible checkmark. Return alone is invisible; a button alone loses the keyboard. Without either, a typed identifier is discarded silently when the field loses focus, which is a worse failure than the button this PR removes.

## The census obligation

`app_rules.add_rule` outlived the Button it used to label. gui.md requires a Settings-row change to update its census entry in the same change set, so the picker that replaced the button now carries that key as its **accessible name**.

Naming a control replaces what VoiceOver derived from the text drawn inside it, so the name gives the value back: the picker announces "Add app rule" *and* which app is chosen, where before it announced only the app. (`AnnouncedValueTests`' walker matches `Picker(`/`Menu {` only, so this Button is outside its census — the obligation is met by hand and pinned by this PR's own guard.)

## Verification

Full gate green: build, `swift test -q` (4318 tests / 820 suites), lint, site build (`docs/` touched). Release build skipped — no concurrency surface.

Worth flagging: **the suite passed unchanged before I added a guard**, which is exactly the "a behavior change owes a test that fails without it" case. The whole change is a wiring change, so the guard is a wiring guard — nothing headless can click a popover row. New `AppRulesAddOnSelectTests` holds three things: the pick reaches a commit, the typed path keeps both commit channels, and the census key is still drawn by whatever replaced the button.

`guard-prover` is running; findings will be posted here and addressed before this leaves draft.

## Not verified here

The feel of it — that picking now lands a row with its icon and nothing flickers between — is an eye check. Cheap to fold into the keyboard-navigation sitting already queued for #996/#991/#997, though it needs no special system setting.

### DIFF
--- a/Sources/KiwiDesk/Settings/Components/Common/AppRuleControls.swift
+++ b/Sources/KiwiDesk/Settings/Components/Common/AppRuleControls.swift
@@ -2,6 +2,12 @@ import KiwiDeskCore
 import SwiftUI
 
 /// App picker with installed apps list and custom bundle ID text entry.
+///
+/// Picking from the list IS the add — there is no separate
+/// confirm, because choosing an app already says everything
+/// (#1172). Free text cannot work that way: every keystroke of
+/// `com.apple.Safari` is a prefix of it, so the typed path keeps
+/// a commit and is the one exception the issue rules.
 struct AppSelector: View {
     /// Bundle identifier of chosen app (`AppRef`).
     @Binding var name: String
@@ -9,6 +15,9 @@ struct AppSelector: View {
     /// already have a rule row, since each app carries at most
     /// one (its space + float facets live on that single row).
     var exclude: Set<String> = []
+    /// Called with the bundle id the user committed. The caller
+    /// owns normalisation and dedup; this view only says when.
+    let onCommit: (String) -> Void
     @State private var custom = false
 
     var body: some View {
@@ -22,6 +31,15 @@ struct AppSelector: View {
                     text: $name
                 )
                 .textFieldStyle(.roundedBorder)
+                .onSubmit(commitCustom)
+                Button(action: commitCustom) {
+                    Image(systemName: "checkmark.circle")
+                }
+                .buttonStyle(.borderless)
+                .disabled(name.trimmed.isEmpty)
+                .iconButtonAffordance(
+                    L("app_rules.add_rule", "Add app rule")
+                )
                 Button {
                     custom = false
                     name = ""
@@ -48,7 +66,10 @@ struct AppSelector: View {
                     : KeybindingCatalog.displayName(
                         forBundleID: name
                     ),
-                onPick: { name = $0.bundleID },
+                onPick: { app in
+                    name = app.bundleID
+                    onCommit(app.bundleID)
+                },
                 escapeLabel: L("app_selector.custom", "Custom…"),
                 onEscape: {
                     custom = true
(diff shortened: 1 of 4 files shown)


hajiboy95:
## `guard-prover` round — 3 red-proofed, 2 real problems found and fixed

All five mutations red with the right message: the dropped `onCommit`, each commit channel removed separately (naming which one was lost), the label removed with the census entry left claiming the key, and the value removed with the label kept.

It also found two things worth having caught before merge, both now fixed and both re-verified by replaying its exact mutations:

### The suite was fail-open on both census clauses

The needles were read file-wide, not against the picker's own chain. The prover proved it twice: move the `.accessibilityLabel` onto the **clear** button in the other branch, and the guard stays green while `app_rules.add_rule` is announced by a button that clears the field and the add affordance has no name at all. Same for the value.

Both clauses now read only the `AppPickerButton(` chain. Replayed the label mutation against the fix: it reds.

### The `pickCommits` needle billed a rename as a regression

It pinned the closure body verbatim, including the local parameter's spelling — so renaming `app` to `picked`, behaviour identical, red the suite with the loudest message in it ("the picker no longer commits what it picked"). A message that is false is worse than no guard, because a reader trusts it and goes hunting.

Anchored now on what the closure cannot lose — a commit reached from inside the `onPick:` argument. Verified: the real regression still reds, the rename stays green.

### Also fixed: the key was hand-restated on both sides

The test typed `"app_rules.add_rule"` as the expected census value *and* inside the scan needle, so the two agreed with each other and with nothing else — renaming the key in `SettingKey+AppRules.swift` would red at the test rather than at the drawing site, and making the test agree with itself would restore green without anything checking the picker draws the new key. The needle is now derived from `AppRulesKey.appRulesAdd.text.label`.

### Noted, not fixed

The suite holds source text, never the rendered accessibility tree — it cannot say what VoiceOver announces. That is the eye/ear check, and it is cheap to fold into the sitting already queued for the focus trio.

Issue #1209, opened 2026-09-02
💡 One command to export the KiwiDesk log for a time frame to a file


hajiboy95:
### Use case

Every bug report wants the log, and today the bug template asks for `log show --predicate 'eventMessage contains[c] "KiwiDesk"'` with no time frame and no destination — the reporter gets thousands of lines in a terminal, has to guess how far back to go, and pastes a fragment. During the #1145 / #1207 sittings the useful reading was always the same shape: "the KiwiDesk subsystem, the last N minutes, written to a file I can attach". A reporter should be able to produce exactly that with one command that answers with the file path.

### Proposed solution

Either a CLI verb or a documented one-liner — decide which, do not ship both:

- **CLI verb**, e.g. `kiwidesk export_log <path> [<last>]` (`last` a duration like `10m`, `2h`, default ~15 min): runs the unified-log query for `subsystem == "com.kiwicanopy.kiwidesk"` (plus the tccd/AX lines a report usually needs?), writes it to `<path>` — a local path, `~` expanded — and prints the resulting file path. Belongs beside `debug_log` (which only WRITES a marker line into the log today; it is not an export). Because the executable is also the app, the verb can run locally without the socket, like `--version` does.
- **Or no verb**: if `/usr/bin/log show` already does the job, the deliverable is the recipe itself, in `docs/cli.md` and the bug template:

  ```sh
  /usr/bin/log show --last 15m \
    --predicate 'subsystem == "com.kiwicanopy.kiwidesk"' \
    --style compact > ~/Desktop/kiwidesk-log.txt
  ```

  (`/usr/bin/log`, spelled out — zsh users often have `log` shadowed.)

In both cases the **bug template** (`.github/ISSUE_TEMPLATE/bug_report.yml` ▸ "Relevant output and logs") is updated in the same change to tell the reporter to run that and attach the file, and to say how far back to reach ("since just before the problem").

### Alternatives considered

- Keep the current template text: it produces unbounded terminal output and fragments, which is what this issue is about.
- A Settings button "Export log…": nicer for GUI users, but the reporters who file issues are already at a terminal, and a verb is the smaller change; the button can come later on top of the verb.
[...]

PR #1211, opened 2026-09-02
docs(cli): export the log for a time frame to a file (#1209)

hajiboy95:
## Summary

Read-only over what KiwiDesk already writes (owner constraint on #1209): the CLI reference gains an *Exporting the Log* section with the unified-log recipe — the subsystem predicate, a time window, output to a file, the live-stream variant — and the bug template's log field points reporters at it (attach the file, reach back to before the problem) instead of the old unbounded `eventMessage contains` filter. Notes that the shipped build's log is readable with no debug build, and that `debug_log` writes a marker and exports nothing.

The Settings **Export Log…** control ruled on #1209 is the next step and will re-point the template at itself when it lands.

## Verification
- `scripts/lint.sh`, `RuleCitationTests`/`InstructionPinTests`/`CiPathFilterTests`, site build: green. Template YAML parses.

Refs #1209

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -72,8 +72,13 @@ body:
         Paste the output of these commands (one per block):
         - `kiwidesk get_state` (print as JSON)
         - `kiwidesk get_layout_info` (current layout details)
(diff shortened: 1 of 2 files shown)


hajiboy95:
Folded into the Export Log… button lane (#1209): the template should point at the GUI once, not at a terminal recipe first. The CLI-reference recipe rides along in that PR as the terminal fallback.

Issue #1207, opened 2026-09-01
🐛 Returning to a Desktop jumps focus to first-in-row (per-Desktop focus memory)


hajiboy95:
### KiwiDesk version or commit

1.1.2 (`main` at 7f93dcea, and reproduced on the #1145 carry branch `feat/1145-sticky-reach-carry` at b77e06b2).

### macOS version

26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. One screen. A KiwiDesk space in a scrolling layout with two windows (say Claude and TextEdit), TextEdit focused and in view.
2. Switch to another macOS Desktop (`focus_desktop 2` or a swipe), then switch back to Desktop 1.

### Expected vs. actual behavior

Expected: Desktop 1 comes back showing the window that was focused when you left it — TextEdit in view, focused.

Actual: on arrival the previously focused window flashes into view, then focus jumps to the FIRST window in the row (Claude), and the scrolling viewport follows it, scrolling TextEdit out of view. Observed by the owner on the 2026-09-01 device sitting and again on 2026-09-02 (the census after returning showed Claude at the viewport origin and TextEdit at x=-1373, scrolled out).

### Relevant output and logs

The mechanism as far as the 2026-09-01 sitting traced it (issue #1145 comment of 2026-09-01):

- The flash-then-jump happens at the SWITCH, not at the 600 ms `desktopSettle`: macOS itself restores the Desktop's focused window, then KiwiDesk's switch handling re-asserts focus from `Space.focused`.
- `Space.focused` is wrong by then because the DEPARTURE ran the close-return cascade: leaving the Desktop makes every window of it vanish, the sweep folds those as destroys, and each destroy re-assigns `Space.focused` down a chain of gone windows — so the raw read is first-in-row exactly when macOS just restored the real one.
- A WIP restore (`rememberDesktopFocus` / `restoredDesktopFocus` on `DesktopMemory`, restoring in the settle) was tried on branch `fix/desktop-switch-reach-and-focus` and does NOT work: (a) it restores in the 600 ms settle while the jump happens at the switch, and (b) the departure cascade has already removed the window from state before the restore runs, so the lookup misses and falls back to first-in-row. It was backed out of the #1145 carry lane by owner ruling (2026-09-01: defer, do not ship half-working).

[...]

PR #1223, opened 2026-09-02
fix(commands): a focus honored while a return's debt stands retires it (#1207)

hajiboy95:
## Description

Follow-up to #1218: CI on #1221 (main plus prose) failed
`OpenOrFocusRingTests ▸ "Re-track after the focus report keeps the
ring"` — #1207's own defect in the other ordering. The switch
handler owed the remembered window, the OS then reported focus on a
returning sibling, and the owed window's arrival paid the debt over
that report. It passed locally only because the host's SkyLight
answers differently for fake window ids.

- `rememberHonoredFocus` retires the debt when a focus is honored in
  the active space while it stands: the report is the OS's or the
  user's own choice, which the memory was only standing in for. The
  payment claims before it raises, so its own echo finds nothing.
- The CI test pins the per-window native Spaces and the topology so
  it reaches the owe on every machine, and reds without the fix
  exactly as CI did. `DesktopFocusPaymentTests ▸
  honoredAfterOweRetiresTheDebt` is the direct clause;
  `ReturningFocusSeamTests` admits the third reader of the debt.
- Rule bullet, §5 row and the design-decisions entry carry the rule.

## Related Issue(s)

Follow-up to #1207 (closed by #1218); unblocks #1220 and #1221,
whose CI hits the same test.

## Checklist

  the app to confirm it works as intended — the mechanism is the
  device-confirmed #1207 one; this ordering is CI's
- [ ] Release build green — CI's `Release Build` job

## How I verified

Full gate: `swift build`, `swift test -q` (green), `scripts/lint.sh`,
site build. Red-proofed by reverting the retire: the pinned CI test
and the new payment clause both red with CI's exact shape
(`lastFocused → window#30`).


hajiboy95:
Review round (code-reviewer, 2026-09-02) addressed in 47cad037: (1) the register tests my marker rewrite had dropped (`theResetForgetsTheMemory`, `theLastDebtIsRetiredFirst`) are restored beside the three-readers clause; (2) the retire no longer fires on KiwiDesk's own raise echoes — `selfEcho` is threaded from the honored tail and gates it, red-proofed by `DesktopFocusPaymentTests ▸ selfEchoDoesNotRetireTheDebt` (stamps the self-raise the way `focusWindow` does); (3) the gate is "rendered in the active space" (`isSticky || space == activeSpace`) so a click on a carried sticky homed elsewhere retires too; (4) the ordering claim about the payer was dropped from the comment and the rule bullet. Advisory items (the test's partial display pinning; pre-existing long prose lines) left as they were.

Issue #1258, opened 2026-09-05
🐛 A stack resize alone in its column refuses silently — no pill, no sound


hajiboy95:
### KiwiDesk version or commit

1.1.2, `feat/resize-refusal-sound` at `02ca5e35` (the #1255
build — found while device-QAing its new pills).

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Put a Space in the **stack** layout with the default
   arrangement (stack zone right, master count 1, so the split
   axis is `x` and the weight axis is `y`).
2. Leave the master zone holding a single window — one window in
   the Space is enough, and a single-master stack of any size
   reaches it too.
3. Focus that window and press **grow height** (or shrink).

### Expected vs. actual behavior

**Expected:** the press is a correct no-op — a column of one has
no share to redistribute — so it cues the way every other
refusal now does since #1255: a pill on the focused window, and
the system alert sound if `refusal.sound` is on.

**Actual:** nothing at all. No pill, no bump, no sound. The
command replies `fail("focused window is alone in its column")`
and the user sees an unchanged screen, which reads as "KiwiDesk
ignored me" — the same complaint that opened #1255.

Reported by the owner during device QA: *"on stack if there's
only 1 window and i go higher it shows nothing"*.

### Relevant output and logs

Not needed — the path is unambiguous in the source, at
`Sources/KiwiDeskCore/Commands/KiwiCore+ResizeStack.swift:88`:

```swift
guard let focusOffset = column.firstIndex(of: focused),
    column.count > 1
else {
    return .fail("focused window is alone in its column")
}
```

A bare `.fail` with no `refuse*` call, so nothing reaches
`cueResizeRefusal` and nothing draws.

### Other window managers or AX tools running?

- [x] None that I know of

### Extra context

**This is not #1255 regressing.** That change replaced the two
*sound-only* cues with pills and made a drawn pill able to
sound; this path was never cued at all, before or after. It
surfaced now only because the pills next to it made its silence
obvious.

**Which cue it earns.** The pill, and not the dead-end
rubber-band. Per #435/#436 the bump is a wordless "nothing
there" reserved for a true no-candidate boundary, while a
*semantic* refusal — the parameter does not exist for this
window — belongs to the pill (`docs/accepted-limitations.md`
states that split). This is semantic: the axis is right, the
column simply has no second member to take the space from.

[...]

Issue #670, opened 2026-08-02
🐛 Native fullscreen: bars stay on top and the settle pass retiles behind the fullscreen app


hajiboy95:
### KiwiDesk version or commit

0.9.4 (`640040ea`)

### macOS version

macOS 26.6

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Enable the Space Bar and App Bar.
2. Put any window into native fullscreen (green traffic-light).
3. Observe both bars still painted on top of the fullscreen
   content, clickable.
4. Wait ~600 ms after the space switch: the previous space's
   focused window receives an AX raise while you sit in
   fullscreen.
5. Retile the home space (open/close a window there): the
   fullscreened window is assigned a layout slot and gets an AX
   frame-set aimed at it.

### Expected vs. actual behavior

Expected: on a fullscreen space KiwiDesk stands down — no bars,
no retile, no focus raise. The fullscreened window is exempt from
layout passes until it leaves fullscreen (it stays a member of
its home virtual space, per the existing design ruling and
`FullscreenStateTests`).

Actual: the concept "fullscreen space" does not exist in the
codebase. Both bars are `NSPanel`s with `.canJoinAllSpaces +
.fullScreenAuxiliary` at `.floating`, `ignoresMouseEvents =
false` (`Bar/BarPanel.swift:31-36`), so they join the fullscreen
space by construction. `handleNativeSpaceChange` no-ops because
`NativeSpaces.activeSpaceNumber()` returns nil on a fullscreen
space (`OS/NativeSpaces.swift:112-120` — indistinguishable from
"SkyLight unavailable"), but `settleAfterNativeSwitch`
(`Profiles/KiwiCore+NativeSpaces.swift:83-106`) still runs 600 ms
later: `retile(force: true)` repaints the bars onto the
fullscreen space and `focusWindow` issues a real AX raise.
`ManagedWindow.isFullscreen` is consulted only by focus rings and
directional focus — never in `Tiling/`, `Layouts/`, or
`Commands/`, so later retiles compute a slot for the fullscreen
window.

### Relevant output and logs

Not captured — the analysis above is static, with file:line
evidence in place of logs.

### Other window managers or AX tools running?

- [ ] Other window managers (e.g., Rectangle, yabai)
- [ ] Accessibility tools (e.g., Keyboard Maestro)
- [x] None that I know of

### Extra context

All hooks for a fix already exist:

[...]

Issue #435, opened 2026-07-22
Explained-refusal pill: generalize + extend to trier, help, i18n


hajiboy95:
## Summary

Today the home-space pill (`sticky.home.pill`, *\"Can only be moved in its home space %1\$@\"*) is the only feedback for a **refused** action, and it fires in just one narrow case. Generalize it into the standard *\"this action is refused, here's why\"* surface and extend coverage.

## Current behavior

- Pill fires **only** on the mouse-drag path, **only** when the *traveler itself* is dragged and snaps back (`KiwiCore+Drag.swift:257` → `flashStickyHomeSpace`).
- Drawn on the dragged window (which in that path *is* the traveler).
- Keyboard `swap` is fully silent — no success check, no feedback (`KiwiCore+NavigateCommand.swift`).

## Scope

1. **Extend to the trier.** When a *different* window is swapped/dropped onto a tiled-sticky traveler, the swap silently no-ops today (`Space.swap` membership guard, `SpaceModel.swift:299`). Fire the pill **on the traveler** — the pill's message is about *its* home-space constraint, so it belongs on the constrained window; the trier just snaps back to its own slot.
2. **`?` contextual help** (#94 affordance) on the sticky shortcut row: *\"A sticky window can only be rearranged in its home space.\"*
3. **Candidate — `track.swap` overflow-fold refusal** (#182): a genuine refusal that is silent today. Same pill vocabulary could explain it. (Evaluate; may split.)

## Explicitly NOT in scope

- Piles are **not** a refusal — swapping with a pile works (drag → pointed member; keyboard from outside → nearest member). Only *keyboard reorder from inside* a pile is intentionally redirected (#172), not blocked. No pill there.
- Wordless *dead-end* feedback (last slot, nothing beyond) is the sibling **shake** issue, not this pill.

## Translation

`sticky.home.pill` is routed through `L()` but the key is missing from `de.json` (English fallback today). Add the German translation during the **next wave's translation session** (via `scripts/merge-keys de`, never hand-edit locales). Any new strings this issue adds join the same session.

## Design principle

Pill = **explained** refusal (semantic, teaches *why*). Distinct from the shake = **wordless** dead-end (spatial). Keep the two vocabularies separate.

hajiboy95:
Sibling: #436 (wordless dead-end shake). This issue = *explained* refusals (pill w/ reason); #436 = *wordless* spatial dead-ends (shake). Kept separate by design.

[...]

Issue #383, opened 2026-07-18
fix(resize): clamp BSP resize to min_window_size instead of collapsing into the overlap pile


hajiboy95:
## What happens

In BSP, resizing a window large enough that a neighbor would fall below `min_window_size` (default 300 pt) makes the windows **suddenly stack** (overlap into an `OverlapStack` title-bar cascade) instead of the resize stopping at the min-size floor.

Reproduces with **both** interactive mouse resize **and** the keyboard grow/shrink shortcuts.

## Mechanism (why it piles)

Both resize paths write a BSP split ratio clamped only to `[0.1, 0.9]` — screen-size- and min-size-blind:
- mouse: `applyResizeAdjustment` `.bspRatioH/.bspRatioV` — `KiwiCore+MouseResizeEnd.swift:104-119`
- keyboard: `resizeBsp` — `KiwiCore+Resize.swift:135-147`

The pure BSP layout then treats "neighbor below `minWindowSize`" not as something to clamp but as a **trigger** to bail into the emergency `OverlapStack` cascade (`BspLayout.swift:58,82` → `OverlapStack.frames`, `OverlapStack.swift:24-49`). Because BSP is a dwindle recursion, a deep sub-region can collapse into the pile at once — hence "suddenly start to stack."

## This is a known, documented limitation

`docs/design-decisions.md:58` already records it: "An extreme stored BSP ratio still collapses the space into the overlap cascade — even though the stack layout no longer does … The stack's layout-time clamp hasn't been migrated to BSP yet … **Migrate the clamp principle to BSP (follow-up to #44).**" This issue is that follow-up.

## Cross-layout picture

- **Stack** — already clamps interactive resize to keep both zones ≥ min (`StackLayout.effectiveRatioRange` / `cappedRatioWrite`, from #44). No longer piles from a resize.
- **BSP** — no resize clamp (this issue).
- **Grid / Track** — also pile via `OverlapStack` when a cell/track falls below min (layout-time only; resize clamp worth considering for parity).
- **Scrolling** — the odd one out: floors the slot at `minWindowSize` instead of piling (`ScrollingLayout.swift:123`).

## Proposed fix

Replace the bare `min(max(value, 0.1), 0.9)` in the two BSP resize write sites with an effective-range clamp analogous to the Stack's `effectiveRatioRange` (`StackLayout+Domain.swift:62`), derived from the region span and `min_window_size` so a resize can't drive a neighbor below the floor.

[...]

PR #1265, opened 2026-09-05
fix(commands): a bsp refusal names the window that cannot move (#1259)

### LINKED ISSUE #1259: 🐛 A bsp refusal names the focused window as its own minimum when a neighbour is the one stuck

1.1.2, `feat/resize-refusal-sound` at `02ca5e35` (the #1255
build, during device QA of its pills).

### macOS version

macOS 26.6.2

### Mac model type

### Steps to reproduce

1. A Space in **bsp** with three or more windows, arranged so
   the focused window spans the whole height — the first split
   is vertical (`x`), the neighbours are stacked on the other
   side and share the horizontal split (`y`).
2. Focus that first, full-height window.
3. Press **shrink height** repeatedly. The first window's own
   height cannot change — it owns the full span — so the press
   moves the neighbours' split instead. That part is correct and
   deliberately silent.
4. Keep going until one of the **neighbours** reaches its
   minimum height.

**Expected:** the window that cannot move gets the pill. #435's
rule, already written into `reportResizeRefusal`'s own
docstring: *"the pill goes on the window that cannot move, not
the trier"*. So the neighbour at its floor should read
**"Minimum window size reached"** and the focused window should
read **"Neighboring window at its minimum size"** — the pair
`refuseGrowAtNeighborMinimum` already draws.

**Actual:** the focused window shows **"Minimum window size
reached"** — the own-minimum wording — even though it is not at
its minimum and its height never changed. The neighbour that
actually hit its floor is not marked at all. The user is told
the wrong window is stuck.

Owner, during device QA: *"when i reach min height of one of the
neighbor windows, it shows min window size has reached in the
first window (since this has focus). only happens when i'm at
the first window which doesn't change on height."*

### Relevant output and logs

Not needed; the path is legible in the source.

`writeCappedBspRatio`
(`Sources/KiwiDeskCore/Commands/KiwiCore+ResizeLimits.swift:216`)
sorts **every** tiled window onto one side of the axis's split
via `MouseResize.bspSide`, then picks the binding side:

```swift
let focusedFirst = firstSide.contains(focused)
let shrinking = deltaSign < 0
let bindingFirst = shrinking ? focusedFirst : !focusedFirst
let binding = bindingFirst ? lowMin : highMin
```

and hands `binding.carrier` to `reportResizeRefusal` (`:111`),
which takes the own-minimum branch whenever the carrier IS the
focused window:

```swift
if let carrier = bindingCarrier, carrier != focused {
    refuseGrowAtNeighborMinimum(focused, anchor: carrier, ...)
} else {
    refuseShrinkAtMinimum(focused, axis: axis)   // ← lands here
}
```

- [x] None that I know of

### Extra context

**The missing premise: participation.** A window spanning the
whole axis being written does not take part in that split at
all — its own size cannot change, whatever the ratio does. The
side classification still files it onto a side, and it can come
back as that side's binding carrier purely because it has the
largest floor there. Nothing downstream asks "can this window's
size change on this axis?", so the trier is handed the
cannot-move wording.

**Not the same bug as the silent cases.** #1258 is a refusal
that draws nothing; this one draws the wrong thing on the wrong
window, which is worse — a wordless no-op reads as "ignored me",
but a confident wrong message sends the user to resize a window
that was never the constraint.

**Where the fix probably belongs.** In `writeCappedBspRatio`, or
in a predicate it consults: a window whose extent on `axis`
equals the layout region's own is not a candidate carrier for
that axis. `reportResizeRefusal` is the shared authority for
stack and track too (`writeCappedMasterRatio`, the track
writers), so check whether the same full-span case is reachable
there before fixing it in one place — the stack master zone
oriented along the split axis is the obvious sibling to test.

**Confirmed working in the same session, for contrast:** the
neighbour-minimum pair itself is right — a genuine neighbour
constraint does put "Minimum window size reached" on the
neighbour and "Neighboring window at its minimum size" on the
focused window. It is only the full-span focused case that
misroutes. The sounds are also correct throughout, including on
sticky/pinned windows.

hajiboy95:
A bsp resize refusal drew "Minimum window size reached" on the
**focused** window when a **neighbour** was the one at its floor
— #435's rule ("the pill goes on the window that cannot move,
not the trier") inverted, and worse than a silent no-op: a
confident wrong message sends the user to resize a window that
was never the constraint.

## The missing premise: participation

A window spanning the whole tiled extent on an axis sits above
every split of that orientation — no move of that ratio can
change its size. The side classification filed it anyway, so it
could come back as its side's binding carrier and wear the
own-minimum pill; its learned floor also entered the **cap**,
blocking a press no window on the split constrained.

Three changes, one rule:

- **`MouseResize.bspSides`** drops those windows, measured
  against the UNION of the slots (the tiling fills its region,
  so no gap arithmetic and no layout context are needed).
- **The binding side is read off the write's own direction**
  rather than the focused window's side, which says nothing
  when that window is on neither. Provably equivalent for a
  participant on both paths — which made `deltaSign` dead, and
  the mouse path's `bspDragSign` with it: that focused-relative
  sign round-trip was how the window's side leaked into the
  decision.
- **`reportResizeRefusal` takes `focusedIsBinding`** in place of
  `shrinking`. The own-minimum wording is owed only to a focused
  window ON the side that could not shrink — one predicate that
  also explains why a grow always takes the neighbour pair, so
  the function got shorter.

## Two cases review opened, both fixed here

- **An axis nothing divides cues on the FIRST press.** The
  first version gated it on the clamp, so early presses said
  nothing while the ratio walked. It is a fact about the
  arrangement rather than a limit reached, so it is owed
  immediately — the stack path answers its own no-parameter
  axis the same way.
- **`noAxisHere` claims the OTHER axis divides**, so it is drawn
  only when that is true. A space dividing neither — one window,
  an overflow pile — would have been told "this zone divides
  widths, not heights" *and* the opposite, both false. It stands
  down wordless instead; that silence is #1258's to close.

**What did NOT change: the write.** Both reviewers read the
ratio write on an unsplit space as an invisible ratchet and
asked for it to be refused before writing. That is wrong, and
the suites say so — refusing it reds twelve cases across
`SessionRatioTests`, `PerSpaceResizeTests` and
`MouseResizeApplyTests`, which drive bsp resize on empty and
one-window spaces on purpose. A ratio is a stored per-space
value whose caps protect the REGION rather than the windows
currently in it (#383/#44/#458), so a later split opens at what
the user set. Only the cue reads the population.

## Stack and track, as the issue asked

Both partition by MEMBERSHIP and every member's extent moves
with the parameter, so the full-span case is unreachable there.
One real consequence, now pinned rather than silent: a focused
window in NEITHER zone (a native-fullscreen focus leaves the
tiled members, #670) no longer reads as at "its own" minimum.
The track's grow branch also now takes the shared door instead
of anchoring the focused window on itself.

## Verification

- `ResizeRefusalTargetingTests` — six arms, bsp and stack.
- `guard-prover`, isolated worktree: four mutations, all four
  red-proofed, plus a fifth I asked for to prove the regression
  pin is not inert. It also caught a dead assertion — a floor
  seeded infeasibly against the span makes the cap give up
  UNCLAMPED, so pre-fix that fixture moved the split and cued
  nothing; re-seeded feasible, with the premise pinned.
- `code-reviewer` + `architect-reviewer`: 4 major, 2 minor, all
  addressed or ruled below.

## Known residue, deliberately not fixed here

- The classification models the FIRST split of an orientation —
  a deeper split reusing the shared scalar is not separately
  classified, so at depth >= 2 a participant can still be named
  on the wrong side. Pre-existing, and the same under-clamp the
  cap already documents.
- `MouseResize` is now an odd home for `bspSides` (no mouse
  caller remains). Moving the shared bsp split geometry out is
  its own chore branch, ruled 2026-09-05.

fixes #1259


hajiboy95:
Device eyeball passed (owner, 2026-09-05), all four arms on a branch build of 951dd527:

1. bsp, three windows, focus the full-height one, shrink height — the pill lands on the stuck NEIGHBOUR, the focused window reads "Neighboring window at its minimum size". The #1259 defect is gone.
2. Two windows side by side, shrink height — "This zone divides widths, not heights" on the FIRST press.
3. One window alone — silent, as designed.
4. A focus that IS in the split still blames itself at its own floor. No overshoot.

Issue #292, opened 2026-07-16
fix(commands): block focused actions while an ignored window is frontmost


hajiboy95:
## Problem

Ignored apps and panels deliberately never enter KiwiDesk's managed
window state. When one becomes frontmost, the active Space therefore
retains its last managed focused window. Commands such as
`toggle_floating`, `resize`, `focus`, or `move_to_space` currently act
on that stale managed window behind the ignored foreground window.

This is dangerous because a shortcut silently mutates a window the user
cannot see.

## Expected behavior

An implicit focused-window command must fail without mutation unless
the OS foreground corresponds to KiwiDesk's currently focused managed
window. Commands that do not target an implicit focused window remain
available.

Return a normal failed `CommandResponse`, for example:

`no managed window is currently focused`

Lua, CLI/IPC, and direct execution must share the same semantics.

## Proposed solution

Add one semantic preflight at `KiwiCore.execute`, backed by a centralized
`FocusedCommandPolicy`. Do not add divergent guards inside individual
commands and do not synchronously query AX focus for every shortcut.

For commands classified as implicit-focused, require:

1. `activeSpace.focused` resolves to a tracked `ManagedWindow`.
2. `NSWorkspace.shared.frontmostApplication` has the same PID.
3. EventLoop still owns observation for that PID.
4. No known ignored/unmanaged panel is focused for that PID.

Fail closed when foreground ownership is nil or uncertain. A transient
rejected shortcut during app activation is safer than mutating a hidden
window.

Block at least:

- `focus`, `swap`
- `move_to_space`, `move_to_space_and_follow`
- `make_floating`, `make_tiled`, `make_auto`, `toggle_floating`
- `resize`, `move_to_track`, `track.swap`
- `stack.promote`, `stack.demote`

Allow:

- `focus_space`
- spawn/pull commands
- layout/config setters and explicit per-Space setters
- profile operations, reload, diagnostics, and state reads
- explicit-ID UI actions such as clicking an App Bar item

Generalize the current Ghostty-specific ignored-panel focus signal into
an unmanaged-window focus latch. It must also cover other filtered
panels, including KiwiDesk's ignored `NSPanel`s. User-ignored apps and
built-in/prohibited apps are blocked naturally because EventLoop owns no
observer for their frontmost PID.

Do not bypass the guard for an outstanding self-raise. Allow once the
OS frontmost PID actually matches; otherwise fail closed until activation
settles.

## Acceptance criteria

[...]

PR #1268, opened 2026-09-05
fix(events): distrust an app's focus of a window we just placed (#1161)

### LINKED ISSUE #1161: 🐛 Focus: Android Emulator self-activation is honored as user focus — the scrolling viewport pans back and the pointer warps

v1.1.1 (installed app; repo at `v1.1.1-4-gc2c6d567`)

### macOS version

macOS 26.6.2

### Mac model type

### Steps to reproduce

1. Scrolling space holding several windows, one of them the
   Android Emulator (standalone window, `qemu-system-aarch64`,
   emulator 37.1.11, AVD Pixel_8_API_36).
2. Focus any other window in the row — the viewport pans away
   from the emulator as expected.
3. Wait a second or two, touching nothing.

Expected: focus and the viewport stay where the user put them.

Actual: ~0.8–1.7 s later, focus snaps back to the emulator —
the viewport pans back to reveal it and the pointer warps onto
it. Reproduced 4× in one sitting (2 during the instrumented
window below, 2 more after).

### Relevant output and logs

The installed build wrote nothing under
`subsystem == "com.kiwicanopy.kiwidesk"` over the repro window
(`log show --last 3h --info --debug` → 0 lines; worth its own
look, the `CoreLog` seam should have narrated the focus
decisions here). The trace instead comes from an external
watcher on `NSWorkspace.didActivateApplicationNotification`
plus an AX observer on the emulator pid, running while the
user reproduced twice:

```
11:51:17.841 ACTIVATED: Antigravity (pid 60714)        ← user focuses away
11:51:19.573 ACTIVATED: qemu-system-aarch64 (pid 9826) ← nobody clicked
11:51:19.948 ACTIVATED: Claude (pid 86199)
11:51:20.332 ACTIVATED: Antigravity (pid 60714)
11:51:20.482 ACTIVATED: Obsidian (pid 1266)
11:51:20.822 ACTIVATED: Tailscale (pid 67207)
11:51:20.957 ACTIVATED: KeyCastr (pid 86830)
11:51:24.157 ACTIVATED: qemu-system-aarch64 (pid 9826)
11:51:24.943 ACTIVATED: Claude (pid 86199)             ← user focuses away
11:51:25.712 ACTIVATED: qemu-system-aarch64 (pid 9826) ← nobody clicked
11:51:28.218 ACTIVATED: Claude (pid 86199)
```

The AX observer on the emulator (registered for
FocusedWindowChanged / Resized / Moved / MainWindowChanged /
ApplicationActivated / WindowCreated) captured **zero**
notifications across the whole window — the only signal is the
workspace-level self-activation.

- [x] Accessibility tools (e.g., Keyboard Maestro)
- [ ] None that I know of

KeyCastr (keystroke display; event-tap only, manages no
windows) and sketchybar (event bridge only).

### Extra context

Working hypothesis (mechanism traced to the handler, fix
direction proposed but not proven):

The emulator's Qt shell re-activates its own app shortly after
losing frontmost — a long-standing qemu-on-macOS misbehavior,
no user input involved (the trace shows no click and no AX
focus report from the app; the activation stands alone).
KiwiDesk's `appActivated`
(`EventLoop+Apps.swift`, "Report the cross-app focus change
ourselves") then synthesizes `.windowFocused` for the
emulator's window, and `handleWindowFocused` honors it: every
distrust machine in that pipeline (ignored-panel dismiss #244,
z-order echo #418/#425, Electron sibling re-report #465/#496)
is keyed to KiwiDesk's *own* raises — an app self-activating
with no raise in flight falls through to the honored path,
which pans the focus-driven layout and warps the pointer
(#186's warp fires precisely because this is "app-driven
focus").

Fix direction: a provenance gate on activation-synthesized
focus — an activation with no recent user input directed at
that app (no click that reached its window, no KiwiDesk
command in flight) is not user focus. The known hard part:
cmd-tab is also a clickless activation and must stay honored,
and the event loop currently has click provenance
(`recentClickReached`) but no keyboard-provenance signal for
system-owned chords, so the gate needs either a listen-only
tap or a narrower heuristic (e.g. distrust only a re-activation
of the app that lost frontmost within the last ~2 s with no
input provenance — a cmd-tab bounce inside that window is the
documented trade). One subsystem, its tests and a device QA
round with the emulator running.

Upstream report to Google's tracker is worth a pointer, but
KiwiDesk surviving misbehaving apps is the durable answer —
the same argument #1157 makes for lazy-AX under-reporting,
from the opposite direction (a spurious event honored, rather
than a missing one trusted).

Workaround meanwhile: run the emulator embedded in Android
Studio (Settings → Tools → Emulator → "Launch in a tool
window") — no standalone qemu window, nothing self-activates.

hajiboy95:
## Description

The Android Emulator self-activation (#1161), diagnosed on
device today and fixed in the same session. The sitting is on
the issue: with the emulator's main window tiled in a scrolling
row and a single `kiwidesk focus left` as the only input, the
emulator focuses itself 0.8–1.5 s after the pan asks it to sit
past the screen edge, 3 of 3 trials; a Space switch that parks it
at the stash corner does the same 1 s later, the window refusing
the corner; and with the window floating, where the pan does not
move it, it never happened. That focus report has a cmd-tab's
shape — clickless, no self-raise stamp, a different window than
the one just reached — and KiwiDesk honored it: the row panned
back, the pointer warped, and from a hidden Space the
focus-follow flew the user back.

The fix is a placement-echo distrust, keyed on where KiwiDesk
itself put the window rather than on activation:

- **`TilingEngine.placements`** (`PlacementLedger`) records each
  window's last placement, stamped at the two frame leaves every
  placer shares (`applyFrame`, `setFrame`), age-bounded at 2 s
  and never consumed — the third echo ledger, and the one read by
  geometry rather than stamp order.
- **`KiwiCore.placementBounce`** decides: a clickless report,
  after focus moved on, for a window in the ledger. In the
  **active scrolling** Space the live entry is the whole verdict —
  a pan that moved the window, on screen or off, or a focus
  command that stepped off it (recorded in the same ledger by
  `focusWindow`, and carried across the pan's later placement).
  That took four device sittings to accept: every narrower
  discriminator this branch tried — past the edge, the size the
  app refused, raised by us, stepped off by a command — honored a
  bounce on the device (the emulator complied within 9 pt and
  bounced; bounced after an on-screen pan; bounced after a step
  off with its bound learned and nothing asked; bounced after a
  click on a neighbour panned the row), and each honored bounce
  left KiwiDesk's focus anchor on a window the OS did not front,
  the split in which the #292 preflight refuses every focus
  shortcut until the app reports again. Anywhere else — the stash
  corner, monocle's park, a scrolling Space you are not on, where
  a clickless focus is how a user *reaches* an off-screen window —
  the placement must lie past the edge and the window must have
  refused it by origin, which the emulator does at the corner
  and a normal window does not.
- **`reassertAgainstPlacementBounce`** keeps state on the window
  the user was on and re-asserts it with a direct raise, the #465
  shape — a placement bounce has no sequence whose closing
  re-assert would put OS focus back, which is why the state-only
  ruling of the raise-echo revert does not reach it. It runs above
  the focus-follow (so the hidden-Space arm stands the follow
  down) and below the z-order revert (a restore's echo keeps its
  state-only revert). A distrust **renews** the placement — the
  emulator retries every half second or so, and its third retry
  landed past a window that started at the pan — through the
  ledger's own `renew` door, which extends the window only while
  the placement itself is younger than it: the report that renews
  is one the predicate cannot tell from a cmd-tab, so an unbounded
  renewal let a user's own retries extend their lockout forever.
  A lockout now ends at most four seconds after KiwiDesk last
  placed the window.

The trade, in `docs/accepted-limitations.md`, priced once: a
cmd-tab onto a scrolling window a pan moved less than two seconds
ago, or back onto the one you stepped off, is bounced for those
seconds; a lockout ends at most four seconds after the last real
placement, your own retries included; a click is always honored.
The discriminator is forgeable by construction, unlike #465's
ordering, and the docs say so.

## Related Issue(s)

Fixes #1161

## Checklist

  (see AGENTS.md §3)
  required)

## How I verified

**Device, on the packaged build of the branch, emulator standalone
(Pixel_3a_API_34), scrolling Space, focus steps over the IPC
socket so the trace carries no user input, a 40 ms CGWindowList
frame sampler beside the log:**

- **First build** (the "not there" clause on both arms): the first
  bounce report was distrusted, the emulator's second report 0.6 s
  later was honored — it had complied within 9 pt of the pan. That
  measurement made the scrolling arm key on the placement alone.
- **Second build, Run K** — main window tiled, three `focus left`
  per trial, two trials: seven bounce reports, all distrusted, none
  honored, the row never panned back. **Run L** — `focus_space 2`
  parking the tiled emulator: no bounce, the active Space stayed 2.
- **Owner, live on the second build:** "much better", then two
  residues — the emulator's third retry landing past the 2 s
  window, and a bounce after a refused size on-screen. Both
  measured in the log and closed by the third batch (renewal on
  distrust, the size arm).
- **Owner, live on the third build:** "if the smartphone is
  offscreen, I go to the very side quickly such that I see it and
  then move away again, then focus jumps back". Reproduced over
  the socket in both directions (19:25): the step onto the
  emulator pans it on-screen and raises it; the step away moves
  nothing; its reports at +0.06 and +0.1 s are our raise's echo,
  the one at +0.75 s is dropped as a self-echo, the one at +1.28 s
  is honored — the self-raise stamp had aged out and both
  geometric arms saw an on-screen, complying placement. The fourth
  batch is that reading.
- **Fourth build:** the relaunch dropped the Accessibility grant
  (52 apps attached, 0 windows tracked); after the owner re-granted
  it, the owner's own quick steps onto the emulator and off again:
  "seems working now". The socket sequence rerun beside that (Run
  K, the quick step both ways, a double quick step) logged eleven
  `placement bounce distrusted` lines and no honored emulator
  report with the socket as the only input — but the owner was
  using Space 1 at the same time (two new windows and a second
  emulator window appeared mid-run), so that trace is a witness,
  not the verification; the eyeball is.
- **Fifth build** (the review batch — the size arm off the learner
  and the bounded renewal), socket only, the owner off the
  keyboard: sitting on the emulator 3.5 s then stepping off, its
  report 0.7 s later was HONORED — the learner had confirmed its
  bound five seconds earlier, so the retile skipped the ask and
  the ledger had no stamp to read; the two further steps were
  distrusted on the edge arm; the quick steps produced no bounce
  this time. That honored report is the displacement batch.
- **Sixth build** (the displacement), the owner's own hands, 21:03:
  bounces distrusted throughout, and three honored — each 0.7 s
  after a pan that moved the emulator ON-screen (a step between
  two other windows, a click onto a neighbour), and a stretch of
  `preflight denied focus` after one of them, the split the owner
  reported as "stuck". That trace is the seventh batch: the
  scrolling arm collapsed to the live entry.
- **Seventh build** (the collapsed arm), the owner's own hands,
  21:17–21:18, an eight-window row with the emulator fifth: steps
  across it both ways, slow and quick, clicks on neighbours that
  panned the row under it — 25 bounce reports, all distrusted,
  no emulator report honored after focus had moved on, no
  preflight denial; "works, tested all of them". Auto-merge armed
  on that eyeball.

Full gate after each batch in the branch's worktree, last on the
merge of `main` (#1259 landed meanwhile and edits the same §5 row):
`swift build`, `swift test -q` (4754 tests, 903 suites),
`scripts/lint.sh` (`KiwiCore+FocusEvents.swift` at 326 lines after
the split), the site build. Release build green locally through
`scripts/build-app.sh`, four times.

Tests: `PlacementBounceTests` — the scrolling arm on a straddling
frame, the compliant-but-past-the-edge case, the on-screen
placement honored, the non-scrolling compliant park honored, the
expired placement, the click escape, the instant and the animated
leaf stamping, the refused size on-screen, the renewal, the
refused stash standing the follow down, the compliant stash
following, gone and rekey; `PlacementBounceRaiseTests` — the late
echo of our raise bounced, past the placement window honored,
without a placement honored, outside scrolling honored, the click
escape, the writer's retention both ways;
`PlacementBounceSeamTests` — the two placement leaves, the
consult, the renewal, the second reading and the writer by exact
count, the raise inside the re-assert over comment-stripped
source. `guard-prover` (isolated worktree, Opus) ran three rounds:
the first batch 7/9, naming the two inert clauses that shaped the
second; the third batch 5/6, the inert one being the refused-size
test whose rect straddled the edge (fixed in the fourth batch:
the rect now sits wholly on screen) plus the finding that the
renewal fixture's near-expiry stamp is what makes its assertion
load-bearing (now said on the stamp); the fourth batch
8/8 red-proofed across ten mutations — the raise arm dropped, the reading widened to the raw ledger and narrowed to the 1 s window, the arm hoisted outside scrolling, the writer pruning at 1 s and never pruning, `raiseWindow` bypassing the writer (the seam row is the sole net there, since `eventLoop.element(for:)` is nil under `makeTestCore`), the placement precondition defaulted away, and a click-provenance vacuity probe.

Review, round one: `code-reviewer`, `architect-reviewer` and
`docs-steward` (Opus) on the first commit — three majors each,
converging on the same three things: the "not where we put it"
clause (wrong in the void, right at the stash), the unmeasured
stash arm, the animated leaf unpinned — plus two false claims in
the row. All addressed in the second commit; the architect's
z-order ordering finding is ruled rather than changed, the reason
in the entry.

Review, round two, the same three on the third and fourth
commits (95372f8e, fc1d31da): 13 majors and 14 minors between
them, converging on four things, all fixed in e10efb0e — the size
arm read a raw frame and so read echo lag as a refusal (now the
learner's answer); the renewal had no end, so a user's own
repeated cmd-tab extended their lockout (now the ledger's bounded
`renew` door); one raw stamp read in the sibling distrust and no
guard on the self ledger's minter (routed, and a write-site
census); and the prose — the row's "a window that went where it
was placed keeps its cmd-tab" had been falsified by the raise
reading, the third discriminator is any focus KiwiDesk itself
gave the window, the active-Space gate was unstated, and the
device narratives had leaked into four docstrings. Dismissed: the
architect's suggestion to gate the size arm on the settled read —
the echo grace is 1 s and the measured on-screen bounce lands at
0.25 s, so that gate would honor the very case the arm exists for;
the learner's candidate tier, seeded only by an echo, is the
reading that keeps the case and drops the lag. Re-review of the
fix batch by `code-reviewer`: all six claims verified; one major on
the size arm's tier fallback (a confirmed width shadowing a
candidate height), moot once the fifth build showed the size arm
was the wrong discriminator and the displacement replaced it; four
minors taken. `guard-prover` on
the fix batch's guards: 5/5 red-proofed, its two findings — a
renew clause the placement clause already implies, and
`updateValue` slipping the write-site census — closed in the next
commit. The displacement batch: `guard-prover` 11/12 red-proofed, the
inert one a ledger test whose chain had a gap wider than the
window (now an unbroken chain), plus two blind spots closed (a
wrapped subscript slipping the census, the fixture rect unpinned).
The collapsed arm: `guard-prover` is running as this lands; its table follows in a comment.

## Notes for Reviewer

This branch was cut from `main` at 59d81161 in a separate
worktree after another session committed its #1259 work onto the
original branch in the shared checkout; both lanes edit the §5
row and the state-and-layout bullet, so whichever lands second
merges main and keeps both fragments.

The logging finding from the issue is closed as a capture-side
error: the installed build streams the subsystem fine with
`/usr/bin/log`; zsh's `log` builtin prints nothing useful.


hajiboy95:
`guard-prover` on the collapsed arm (f538ddca), isolated worktree, Opus:

| Mutation | Red |
|---|---|
| scrolling arm narrowed back to the edge | the flipped on-screen test, the displacement test, the carried-displacement test |
| scrolling arm dropped | those plus the compliant-past-the-edge test |
| active-Space gate dropped | **nothing** — the stash fixture's hidden Space defaulted to bsp, so the mode gate covered it; closed in the next commit by minting that Space scrolling, which turns the compliant-stash test into the guard (proven both ways by the prover before the change) |
| mode gate dropped | the monocle park test |
| click escape removed | every click-escape test in three suites, and the mid-pan test's two assertions — the click added there is load-bearing |
| a second `stampSelfRaise(` site | the seam row |
| a bare `selfRaiseStamps[id] = now` | the write-site census |

4 red-proofed, 1 inert (closed), 0 vacuous. Fixture note taken too: the bounce suite now pins its on-screen rect inside the pinned bounds like its displacement sibling.

Issue #1177, opened 2026-08-31
💡 Gather out-of-bounds windows when a space switches to floating


hajiboy95:
### Use case

Switching a space's layout to floating from scrolling or
monocle can leave windows partly or fully outside the visible
bounds — scrolling's scrolled-out frames, monocle's parked
frames — where they are hard or impossible to reach with the
mouse. Floating keeps whatever geometry the previous mode left,
so the user inherits an unusable arrangement.

### Proposed solution

On entering floating mode, any window **partly or fully**
outside the visible bounds is brought inside — gathered toward
the middle of the screen, in the spirit of the quit-time
teardown gather (`Sources/KiwiDeskCore/App/KiwiCore+Teardown.swift`,
`WindowGather`) but scoped to the switching space and including
its windows regardless of float flags. Windows already fully
visible stay exactly where they are.

Scope by **visibility, not by the previous mode**: a plain
tiled→floating switch never trips it (everything was visible),
scrolling and monocle are covered automatically, and the rule
stays predictable without a mode matrix.

### Alternatives considered

- Only fully-outside windows — ruled 2026-08-31: partly-outside
  count too; a window with a sliver on-screen is still
  effectively unreachable.
- Triggering only for scrolling/monocle origins — the
  visibility test generalizes and needs no origin list.

### Extra context

The gathered frames still respect the bar strips (the float
clamp machinery in `KiwiCore+FloatClamp.swift`), and its
sibling bug about floating-mode spaces not being clamped at all
is filed separately alongside this issue.

hajiboy95:
The sibling bug this body calls "filed separately alongside" — floating-mode spaces never clamped clear of the bars — is #1178.

hajiboy95:
Owner ruling 2026-09-09: this is crucial and goes wrong often — switching a space from monocle or scrolling to floating loses a lot of windows behind the visible ones or off-screen. Priority raised Low → High and the lane moves up in the 1.3.0 order, right after Phase 0b. The parked-corner half of the same symptom is #1352 (a float whose capture was lost is re-centred by a retile-time net); this issue keeps the frames outside the visible bounds that are NOT at the stash corner — scrolled-out and monocle-parked ones — and the proposed `WindowGather` shape stands.

hajiboy95:
[...]

PR #1361, opened 2026-09-09
fix(tiling): re-centre a float stranded at the stash corner

### LINKED ISSUE #1352: 🐛 A floating window comes back at the stash corner: its remembered position is lost

1.2.1 (4c7608a)

### macOS version

macOS 26.6.2

### Mac model type

### Steps to reproduce

Not reliably reproducible from the user's side — it happens "sometimes" over a day of use, on a two-display setup with several Desktops and two profiles. Reading the code and measuring the device gives four independent ways in, all ending in the same state:

1. Put a window on a floating-mode space (or float-flag a window), switch to another space so it parks at the stash corner, then hit one of:
   - **A late park echo from a lifted window** (measured 2026-09-09, see logs): macOS lifts a parked window off the line it was asked for — 4 pt on Safari, Finder and Keyty, 18 pt on a tall window. `TilingEngine.looksStashed` allowed 2 pt, so a slow app's echo landing past the applier's 1 s grace read as a *user move*, `forgetStash` consumed the capture while the window still sat at the corner, and the next forced stash captured the CORNER as the original.
   - **A Desktop switch**: the window `vanished`, the capture is swept with it (`restoreStashed` sweeps entries for windows gone from state), and on return the window is re-tracked at the corner and the next stash captures that.
   - **A relaunch / wake**: the session snapshot records the STATE frame, which for a parked float is the corner; `KiwiCore.restore` replays it verbatim with no capture.
   - **A profile switch** that gives a space a different mode: windows parked without a capture (tiled at park time) become floating-mode members whose only frame is the corner.
2. Switch back to the floating space.

Expected: the window comes back where it was. If the remembered position is genuinely gone, it comes back somewhere usable — centred on its screen, clear of the bars — never at the corner.

Actual: the window stays at the bottom corner with a 1 pt sliver showing, exactly as it looks while another space is open, and every later activation puts it back there because the corner is now its "original".

### Relevant output and logs

Device measurement 2026-09-09 (CGWindowList, on-screen windows on inactive spaces; `dy` is the distance from the park line `bounds.maxY - stashPeekY`, `dx` from the corner x):

```
51   Zen     (-1715,1069,1716,1009) dy=+0  dxL=+0
2754 Code    (-1715,1069,1716,1009) dy=+0  dxL=+0
475  Safari  (-1715,1065,1716,1009) dy=-4  dxL=+0
529  Safari  (-1715,1065,1716,1009) dy=-4  dxL=+0
225  Keyty   (-792,1065,793,536)    dy=-4  dxL=+0
5356 Finder  (4287,1229,2548,1398)  dy=-4  dxR=+0
2779 KiwiDesk(4287,1215,2555,1400)  dy=-18 dxR=+0
```

Five of seven parked windows sit outside the 2 pt tolerance the corner test allowed. `kiwidesk get_state` carries no frames, so the table above is the state; the log has no line for the capture drop (there is none to export).

- [x] None that I know of

### Extra context

Fix direction, one construction rule rather than four patches: **a corner is never an original.** `stash` refuses to capture a frame that looks parked; `looksStashed` tolerates the measured OS lift (one `visibilityFloor`, derived rather than restated); a retile-time recovery (`KiwiCore.recoverStrandedFloats`) seeds a centred frame inside `floatBounds` for any effective-float window on a shown space that has no capture and sits at a corner, so the existing restore pass delivers it; and the session snapshot records the capture instead of the parked frame, so a relaunch puts the window back where it was rather than centring it.

Sibling: #1177 (windows scrolled out or parked by monocle when a space *switches* to floating) — that lane covers frames outside the visible bounds that are not at the stash corner; this one covers the corner.

hajiboy95:
## Description

A floating window sometimes came back at the bottom corner where hidden spaces park their windows, and every later activation put it back there. Its remembered position (the stash capture) could be lost while it still sat at the corner, after which the next park captured the corner itself as the "original".

Four ways in were traced or measured; the fix is one rule instead of four patches — **a corner is never an original**:

- `TilingEngine.looksStashed` gained a pure, bounds-taking variant whose y tolerance is one `visibilityFloor`. macOS lifts a parked window off the line it was asked for (measured 4 pt on Safari, Finder and Keyty, 18 pt on a tall window); with the old 2 pt tolerance a slow app's late park echo read as a user move and consumed the capture.
- `stash` refuses to capture a frame that looks parked.
- `KiwiCore.recoverStrandedFloats` (new, `App/KiwiCore+FloatRecovery.swift`) runs ahead of every retile and seeds a centred capture inside `floatBounds` for an effective float on a shown space that has no capture and sits at a corner; the retile's own restore pass delivers it — one delivery path.
- `sessionSnapshot()` writes a parked float's capture into the crash/sleep snapshot instead of the parked state frame, so a relaunch puts the window back where it was rather than in the middle.

Docs: user guide (Floating), a design-decisions entry arguing the rule, a `state-and-layout.md` bullet and the AGENTS.md §5 row clause.

## Related Issue(s)

Fixes #1352

Sibling: #1177 keeps the frames outside the visible bounds that are NOT at the corner (scrolled-out, monocle-parked) and must not add a second centring.

## Checklist

  (see AGENTS.md §3)
  required)

## How I verified

- Device measurement of the mechanism on the running 1.2.1 (CGWindowList over the parked windows of inactive spaces): five of seven parked windows sat 4–18 pt above the park line, outside the 2 pt tolerance — the table is on #1352.
- `swift build`, `scripts/lint.sh`, the stash-related suites (59 tests in 12 suites, including `RuleCitationTests` and both bounds-routing guards) green; the full run and a `guard-prover` pass over the eight mutations are recorded on the thread below.
- **Owner device eyeball still owed** (not run here, since it displaces the installed app): float a window on a two-display setup, switch away and back a few times including a Desktop switch and a relaunch, and confirm it returns to its place — and that a float found at the corner comes back centred rather than staying there. Auto-merge is not armed until then.

## Notes for Reviewer

The recovery is a pure decision (seed the capture) so `restoreStashed` stays the one writer of frames on that path; the bar clamp after the retile reads the stale corner frame for one pass exactly as it does for every restore today, and the settle retile converges it. `StashCornerLiftTests` pins the lift band by derivation (`visibilityFloor`), never the measured numbers.


hajiboy95:
Review round (code-reviewer + architect-reviewer on be6d469), dispositions in 8020a11c / a9e86c3:

**Fixed**
- code-reviewer major ×2 — `restore` replayed the substituted capture as a frame set that the next forced park overwrote (and at boot the scan's retile had already seeded a centred capture that outranked it), so a relaunch still centred. It now SEEDS the capture for a window sitting at a corner and leaves a record that is itself a corner alone; the boot shape is pinned in `FloatStrandSeamTests` ▸ *A restore seeds a parked float's record as its capture*.
- architect major — the `captureState → sessionSnapshot` wiring was unpinned; *Both captureState seams take the session snapshot* reads both closures through the bootstrap wiring.
- architect minor — one corner predicate: the static `looksStashed(_:)` over `NSScreen.screens` is gone; `TilingEngine.looksStashed(_:)` is an instance method over the `allScreenBounds` seam, asked by the echo classifier and the recovery alike, so the lifted-echo path is now pinned (*A lifted park echo keeps the capture; a real move drops it*). The bounds-routing census entry for the file drops from 2 to 1.
- architect minor (Desktop sweep, taken now rather than as a follow-up because the owner asked whether windows get their PLACE back) — the restore sweep spares every id the away ledger knows, so a window that left with its Desktop returns to its place; the recovery stays the net beneath a memory that is genuinely gone.
- code-reviewer minor — sticky exclusion now asks the park's own `stickyExemptFromStash` (a display sticky parks off its home display and rides the same capture).
- code-reviewer minor — comment diet: the four-path history is in the design-decisions entry only; the measurement lives in the `looksStashed(_:in:)` docstring alone, cited from the rule file and the design entry.
- code-reviewer minor — the user-guide sentence carries `:::unreleased`.
- code-reviewer minor — the decision clauses call `recoverStrandedFloats()` directly; the suite doc names the unpinned display-gone read the two `retile()` clauses rely on.

**Dismissed**
- architect Q3 (move the substitution into `StateCoordinator.snapshot()`): agreed with the architect's own answer — State must not depend on a Tiling ledger.
- `ResetEscapeHatchTests` hand-wires `captureState = state.snapshot()`: it tests the reset hatch's store, not the session's contents; the wiring clause reads the bootstrap wiring, which is the obligation.

Full run on 8020a11c: 5024 tests in 964 suites, the one red being the routing census count fixed in a9e86c3. Guard-prover on the round-2 clauses is running; its table follows.

hajiboy95:
Guard-prover on 8020a11c (isolated worktree, all 11 screen-gated cases ran): **16 clauses red-proofed** over 17 mutations — the lift band both ways and the exact-x match, the corner-capture refusal, the retile wiring, the effective-float / pending-capture / corner gates of the recovery, `sessionSnapshot` and BOTH `captureState` wirings (each seam alone), the away-ledger exemption in the sweep, the restore's seed-not-set shape and its corner-record refusal, the echo classifier's corner test, and the centring arithmetic. One clause was uncovered: the sticky exemption in `recoverStrandedFloats` — added in e0927329 (*What the park exempts, the recovery exempts*), proved red by hand under the same mutation (a global sticky at the corner came back centred at (560, 252.5)) and green restored.

Full run on the round-2 tree: 5024 tests in 964 suites; CI green on a9e86c3, e0927329 is the test-only addition.

Owner eyeball still owed before `--auto` (two displays: switch away and back, a Desktop switch, a relaunch — the float returns to its PLACE; a float found at the corner comes back centred).

Issue #1395, opened 2026-09-13
📝 Streamline the user guide under concise-phrasing rules, and add those rules to AGENTS.md


hajiboy95:
### Which doc?

`docs/user-guide.md` (3,472 lines) — whole document, plus `AGENTS.md` §2 for the accompanying rule.

### What kind of problem?

Structure — hard to find / poorly organized

### What's wrong or missing?

The user guide re-explains settings that the Settings UI already explains in its own row captions and tooltips, and it argues rulings that belong in `docs/design-decisions.md`. A reader looking for "what does this switch do" wades through paragraphs the UI already says in a line. The guide should be the concise map, and the arguments should live where docs.md already says they live.

### Suggested fix (optional)

Action plan, in order:

1. **Apply concise phrasing to the user guide first.** Remove explanations already carried by the Settings UI and its tooltips; keep what the UI cannot say (cross-cutting behaviour, keyboard flows, limits). Base the phrasing rules on the Caveman guidelines — https://claudemarketplaces.com/skills/juliusbrussee/cavekit/caveman — adapted for prose that stays readable by a user: short sentences, one fact each, no restated rationale.
2. **Add the global comment and doc rules to `AGENTS.md`** (§2 for comments, docs.md for prose) rather than into individual subagent definitions, so every future review applies them. §2.8 already states the comment budget; the doc half is new.
3. **Audit the remaining docs** under collector #1406, one issue per file, Medium priority and no milestone.

Milestone 1.4.0 for this issue (the user guide and the rule).

Issue #1018, opened 2026-08-25
💡 A fresh install should open in Scrolling on every screen but the smallest


hajiboy95:
### Use case

The first Space a fresh install opens in decides whether tiling
feels good or feels like the app squashed everything you had.

Today that layout comes from `ScreenClass.layouts`, best-first
per screen shape: a 2560 pt desktop leads with **Grid**, an
ultrawide with **Track**, a pivoted screen with **Stack**. Only
the laptop class leads with Scrolling. So the first thing most
users see is their windows cut into halves or thirds — the exact
impression that makes people close a tiling manager on day one.

Scrolling is the mode where nothing gets squashed: each window
keeps a comfortable slot and the neighbours wait one keystroke
away. It should be what a new user meets first.

### Proposed solution

Owner ruling, 2026-08-25. The rule is by **size**, not by which
screen is main:

- Rank the connected screens by width. The **smallest** screen
  leads **Monocle**; **every other screen leads Scrolling**.
- With one screen only, that screen leads Scrolling — at least
  one Scrolling space always exists. A solo laptop and a solo
  27" both open in Scrolling.
- Three screens: the smallest leads Monocle, the other two lead
  Scrolling.
- Behind that first slot, each screen's remaining spaces keep
  today's per-class ordering. The one-Floating-space rule and
  the budget/share arithmetic in `StarterAllocation` are
  unchanged.

**Slot size — one profile-wide value, no per-space overrides**
(owner: *"we keep it one value … no need to use override then"*).
`StarterTuning` sets `scrolling.slotSize` explicitly instead of
leaving it on `.auto`:

- **48 %** as the shipped starter value. Two columns side by
  side, and it reads well on a laptop as well as on a 27" — where
  today's `.auto` resolves to 95 %, a near-full slot with a
  sliver peeking.
- An **ultrawide** main screen may want ~30 % (three columns).
  That is the owner's open suggestion, not a settled number —
  rule it during implementation.

### Worked examples

Derived from the allocator as it stands (`budget` → `shares` →
`floatingHost` → `take`), with the new first-slot rule on top.
One screen gets a budget of 3 and is its own Floating host, so
its quota is two tiled spaces plus Floating.

| Screens | Today | Under this rule |
|---|---|---|
| Laptop alone (1728 pt) | Scrolling, Monocle, Floating | **unchanged** |
| 27" alone (2560 pt) | Grid, Stack, Floating | Scrolling, Grid, Floating |
[...]

Issue #634, opened 2026-07-29
Settings: two-tier reset escape hatch — discard saved arrangement + Reset All Settings


hajiboy95:
## Summary

When restore/config state goes bad (see #633) there is no
in-app escape hatch — recovering means knowing about
`~/.config/KiwiDesk/` and hand-deleting files. Add reset
affordances. Shaped by a ui-designer consult (2026-07-29) against
the §2.7 north star; two tiers, not one nuclear button.

## Tier 1 — "Discard Saved Window Arrangement" (ship first)

- **Where:** Behavior tab, directly beside the wake-restore
  controls (`enable_wake_restore` / delay) — the escape hatch
  sits with the setting it undoes.
- **What:** deletes `.state_snapshot` + `.session_snapshot` (and
  drops any in-memory held wake snapshot).
- **Presentation:** plain `.bordered` row button, caption
  "Clears the arrangement KiwiDesk remembered from your last
  session or wake, without changing any other settings."
- **No confirmation dialog:** strictly less consequential than
  today's unconfirmed single-profile delete — the files
  regenerate within one 30 s autosave cycle. Independent of the
  dashboard's `isDirty`/discard-gate machinery (the snapshots are
  not part of the edited `GuiConfig`).

Directly targets the actual bug reports (bad session-snapshot
restore) at near-zero risk.

## Tier 2 — "Reset All Settings…"

- **Where:** General ▸ Advanced disclosure, **last** row, below
  "Edit init.lua directly", separated by a second `Divider()` —
  ascending-severity ladder: Reveal (read-only) → Edit init.lua
  (mode switch) → Reset (wipe).
- **Label:** "Reset All Settings…" — mirrors iOS Settings ▸
  General ▸ Transfer or Reset ▸ "Reset All Settings" (config
  wiped, content kept). NOT "Total reset"/"Factory reset": both
  over-promise once init.lua visibly survives. Trailing ellipsis
  per HIG (command opens a confirmation).
- **Control:** plain `.bordered` `Label` with
  `arrow.counterclockwise` — danger is signaled by the dialog's
  destructive role, not a resting red button (house convention).
- **Confirmation:** its own `confirmationDialog` (never stacked
  on the shared `DiscardConfirmation` gate — that only fires
  `if isDirty`; this must confirm every time):
  - Title: "Reset all KiwiDesk settings?"
  - Message: "This deletes every saved profile, your spaces,
    layouts, and keybindings, and forgets any window arrangement
[...]

PR #1454, opened 2026-09-14
fix(profiles): heal a screen the resolve left with no space

### LINKED ISSUE #1175: 🐛 Pinning spaces via Lua can leave a Desktop with zero spaces

1.1.1 (current `main` on the dev machine)

### macOS version

macOS 26.6.2

### Mac model type

### Steps to reproduce

1. In the Lua config, pin spaces to Desktops such that every
   space that would live on some Desktop is pinned elsewhere
2. Apply the config
3. Switch to the emptied Desktop

Expected: every Desktop always carries at least one space —
there is always somewhere for windows to land and something for
the bars and space verbs to address.

Actual: the pinning is accepted as written and a Desktop can end
up with zero spaces.

### Relevant output and logs

Not captured — this is a config-shape condition, reproducible
from the config alone; no runtime log is needed to see the
mechanism.

- [x] None that I know of

### Extra context

Owner ruling 2026-08-31: **heal, don't refuse.** A Desktop that
would end up empty auto-seeds a default space instead of the
config being rejected — refusal was considered and ruled out
because it makes pin order matter and turns a valid-looking
config into an error. Healing also matches how the starter setup
already derives spaces from what is connected.

hajiboy95:
## Description

A connected screen the space→display resolve leaves with no space is healed, never refused (#1175, owner ruling 2026-08-31). `init.lua` could pin a screen's last space away, and a profile of named spaces saved for one screen loaded dirty onto two left the second with nothing to show.

- **One heal where every relocation ends.** `healEmptyDisplays` runs at the tail of `resolveSpaceDisplays` — the total resolve the Lua pin, every profile apply, the GUI apply, a monitor change and a space delete all funnel through — and is called by `move_space_to_display`, the one verb that relocates beside the resolve on purpose. It seeds one numbered space (smallest free number) in the starter setup's lead layout for that screen (`StarterAllocation`, #1018), pinned to the monitor so the next resolve keeps it, and logs `heal: display 'B' had no space — seeded space N`.
- **A fingerprint-keyed ledger** (`healedSpaces`, session state beside `spacePins`) makes it idempotent: a pin reset that did not prune — a re-dock onto the live profile's own set, a config reload — re-pins the earlier seed rather than minting one per event. Every apply door (profile, composed Standard, GUI draft) retires the seeds its declared set adopts, so a seed a Keep or a Settings Save captured is the user's and the heal seeds *beside* it. Twins (one fingerprint, two screens — the standing pin limitation) mint once for the pair. The #634 reset clears the ledger.
- The heal runs **before** relocation is judged for the float re-anchor, so a reused seed the precedence sent to main and the heal sent back is not re-anchored twice.
- `SecondarySwitchTests` asserted a whole-map emptiness that rested on a display with no space; it now pins the main Desktop's memory per key. `commandSizing`'s docstring is cut to §2.8 budget to make room for the ledger in a file at the 350 ceiling.
- Docs: a design-decisions entry (the ruling, the rejected refusal, the accepted re-dock residue), the Lua reference states the heal once under *Profile Monitor Sets* and links from `pin_space_to_display` and `delete_space` — all `:::unreleased` — and profiles.md gains the section plus the §5 clause.

Not added, per the #1395 ruling: a line in `spaces-and-desktops.md` ("Every Space sits on a screen" now has a true converse). docs-steward flagged it as the owner's call — say the word and it's one clause.

## Related Issue(s)

Fixes #1175

## Checklist

## How I verified

- verify-gate: `swift build`, `swift test -q` (5211 tests / 998 suites green), `scripts/lint.sh` exit 0, `scripts/unreleased-strip --check`, site build green; release build left to CI.
- guard-prover, three rounds, all red-proofed: the heal call removed, the pin dropped, the ledger reuse disabled, the lead mode dropped, the id minting broken, the move-door call removed, the retire made a no-op, the twin arm collapsed both ways, a stray ledger writer, the heal moved after the re-anchor, the composed door's retire removed — each redding the test that names it. Round 1 found the twin fixture hash-order vacuous; it now pins both spaces to the shared fingerprint and asserts the heal fired.
- Reviews: code-reviewer + architect-reviewer + docs-steward (round 1: 1 blocker, 10 major, 7 minor — all addressed), sequential code-reviewer re-review (1 major — the composed door — addressed). Blind spots that remain and are stated: the GUI door's retire is held by the census, not a behavioural fixture; the float strand is held by the order clause, since no unit fixture can observe the re-anchor (it resolves real `NSScreen`s).
- **Not run on the device**: the heal fires only with two screens connected, and this machine has one today. **Owner eyeball owed before merge** (two screens, "Displays have separate Spaces" on): `pin_space_to_display` every space of screen B onto A and watch B get a seeded space in the Space Bar; then `load_profile` a one-screen profile of named spaces; then a re-dock onto the live profile's own set should not add a second seed. Auto-merge is deliberately NOT armed.

## Notes for Reviewer

`ProfileComposition.compose` already seeds a monocle space per screen beyond a Standard's count — a sibling concern (a composed Standard covering more screens than it plans), not the same mechanism; the architect noted it as a pre-existing second answer to "what layout does an uncovered screen open in" and not this PR's to restructure.


hajiboy95:
**Device sitting, 2026-09-14** — built-in 3456×2234 (main) + DELL U2720Q, "Displays have separate Spaces" on, this branch's signed build launched from scratch (62 apps attached, grant intact), profile `My Setup` (1–2 Main-role on the built-in, 3–5 pinned to the DELL). Log filtered by pid.

| Door | Drove | Saw |
|---|---|---|
| Lua pin | `pin_space_to_display 3/4/5 → 1` | exactly one `heal: display 'DELL U2720Q' had no space — seeded space 6`, on the third pin; `6` in Scrolling; the DELL's Space Bar showed only `6` |
| Window on the seed | `move_to_space 6` | tiled full-frame on the DELL, bars followed; moved back |
| No over-eager heal | re-pin 3–5 to the DELL, `delete_space 6` | zero heal lines, DELL back to `3 4 5` |
| `move_space_to_display` | move 3/4/5 → 1 | its own heal fired once on the third move, seeded `6` |
| Restore via resolve | `delete_space 6` | zero heal lines (the resolve reverted the runtime moves), modes and windows untouched |
| Profile door | `load_profile HealProbe` (one screen, named `work`/`mail`) | dirty load, `work`/`mail` on the built-in, DELL healed with `1` (smallest free number after the prune) in Scrolling |
| Prune and reseed | `load_profile HealProbe` again | one heal line, still one seed |
| Back | `load_profile "My Setup"` | 1–5 back, zero heal lines |

Auto-merge armed after the owner's go.

Issue #818, opened 2026-08-11
🐛 A help string that quotes another key's label drifts from it per locale


hajiboy95:
### KiwiDesk version or commit

0.9.5 (64bb40fa)

### macOS version

macOS 26.6.1

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

Found by review rather than by use, so the repro is a read of
the shipped catalogs rather than a click path. Set the GUI
language to German (Settings ▸ General ▸ Language), then:

1. Go to Layout Defaults ▸ Track.
2. Open the `?` popover on **New window position**.
3. Compare the option names the popover quotes with the names in
   the menu beside it.
4. Repeat with the language set to Japanese.

Or read it straight out of the catalogs:

```
python3 -c "
import json
for loc in ['en','de','ja']:
    d=json.load(open(f'Sources/KiwiDeskCore/Resources/Locales/{loc}.json'))
    print(loc, repr(d['track.new_window.own']))
    print('  ', repr(d['track.new_window_position.help'])[:120])
"
```

### Expected vs. actual behavior

**Expected:** where a help string quotes an option's name, the
quoted words are the words on the control, in every language.

**Actual, `de`:** the option reads "Öffnet seinen eigenen Track";
the help quotes „Öffnet eigenen Track". Close enough to look
right, different enough that the help names something not on
screen.

**Actual, `ja`:** the help is a one-line summary
(「新しく開くトラックまたはウインドウの挿入位置を指定します。」)
that quotes neither option, so the English's "with X …, with Y …"
mapping — the whole point of the sentence — is gone.

The English is fine, and that is the trap: the label and its
quotation were typed once, by one person, in one place. Every
locale after that holds them as two independent strings that must
agree forever, with nothing checking that they do. It is a
hand-kept mirror, which AGENTS.md §5 (`parity-tests.md`) says
needs a forget-proof test — and this one has none.

`track.new_window_position.help` is the instance found; the class
is any help or caption that quotes another key's value.

### Relevant output and logs

Not applicable — a catalog-content defect, nothing at runtime to
log.

### Other window managers or AX tools running?

- [x] None that I know of

### Extra context

Found during the #678 Phase 4 pass 9+10 review (PR #817), where
the same class was nearly shipped again: `onboarding.grant.body`
quotes the **Open System Settings** button label in its numbered
steps. That pair is consistent in all 11 locales today, and the
two keys behind it were merged into one
[...]

PR #1455, opened 2026-09-14
fix(gui): the greyed Focused window row links its reason beneath the grid

### LINKED ISSUE #1310: 🐛 A greyed Advanced Colors row shows no reason, and its tooltip never appears

1.1.2 — build of `origin/main` (`a7ec5989`).

### macOS version

macOS 26.6.2 (25G83)

### Mac model type

### Steps to reproduce

1. Have the Space Bar's **App symbol style** set to anything other
   than **Glyphs** (the shipped starter setup does), so the
   `focusedItemInert` gate is live.
2. Open Settings ▸ **Advanced Colors** ▸ *Space Bar colors*.
3. Look at the **Focused window** row: it is greyed and inert.
4. Hover the pointer over the greyed row and wait several seconds.

**Expected:** the user can find out *why* the row is inert without
guessing. AGENTS.md §5's GUI row is "grey don't hide", and the
sentence that explains this particular grey already exists —
`space_bar.color.focused_item.help`: "Tints the focused window …
Glyph tint needs "App symbol style" set to "Glyphs"."

**Actual:** the row is greyed with no reason visible anywhere, and
**hovering it produces no tooltip at all** (owner, 2026-09-07). The
explanation is written, localized and wired, and the user cannot
reach it by any route.

Reported by the owner during the #1309 sitting: "now it's
deactivated and unclear why".

### Relevant output and logs

Two separate problems, and the first is the one that makes the
second invisible.

**1. A row-level gate has no live anchor, by design.**
`GreyOut`'s own docstring already rules that its `help:` is "a
HOVER fallback only, never the #94 affordance", because
`.disabled` is cumulative and so a `HelpButton` *inside* the gated
content is dead — "a block gate anchors a live `?` outside the
subtree (#527)". The Borders group does exactly that, via
`AdvancedColorsGates.bordersHeaderHelp` on the `SettingsSection`
header. The Space Bar card's *row*-level gate
(`AdvancedColorRow+Bars.swift:102`, `gated(gates.focusedItemInert,
focusedItemHelp)`) has no such anchor — the convention as written
covers block gates only, so a row gate's reason has nowhere live
to go.

**2. Leading hypothesis for "hover shows nothing": an inactive
outer `GreyOut` stamps `.help("")` over the row.**

`GreyOut` writes its tooltip unconditionally:

```swift
content
    .disabled(active)
    .opacity(active && !alreadyDimmed ? 0.5 : 1)
    .help(active ? help : "")          // <- "" when INACTIVE
```

and `AdvancedColorRows` wraps *every* Advanced Colors row in an
outer `GreyOut` for the container gate:

```swift
AdvancedColorRow(...)
    .modifier(GreyOut(active: gate.containerGrey, help: gateHelp))
```

When the Space Bar is ON, `containerGrey` is `false`, so that outer
modifier applies `.help("")` on top of the row that has just set
its own `.help(focusedItemHelp)` inside. An ancestor tooltip
covering the same region plausibly shadows the descendant's, which
would make the row's explanation unreachable in exactly the state
where it is the only explanation there is.

That is a hypothesis, not a measurement. Two cheap checks decide
it:

- make `GreyOut` apply no `.help` at all when inactive
  (`if active { $0.help(help) }`) and hover the same row;
- or hover a row whose only `GreyOut` is the active one, and
  compare.

Note the standing counter-ruling recorded in `GreyOut`'s docstring:
"a 'no tooltip' report is AppKit's default tooltip delay until
proven otherwise — `.disabled()` does NOT suppress `.help()`
(2026-08-03)". The owner did wait on the hover, so that ruling is
worth re-measuring here rather than assumed; if it holds, the
shadowing hypothesis above is the remaining candidate.

**Scope to determine before choosing a fix.** Every row-level gate
routed through `AdvancedColorRow.gated(_:_:)` is in the same
position, not just this one — so the repair belongs at the gate
delivery, not at this row. Whether the answer is a live `?`
anchored beside the row's label (outside the disabled subtree), a
caption under the card, or fixing the shadowed tooltip is a
`ui-designer` question; `docs/ui-patterns.md` and #527 own the
existing conventions.

- [x] None that I know of

### Extra context

Found while diagnosing #1309 on the same page, but independent of
it — this is about the gate's explanation, not about focus.

The sentence itself is already correct and already interpolates
the picker's own key rather than quoting it (#818), so no
localization work is implied by fixing the delivery.

hajiboy95:
## Description

On Advanced Colors ▸ Space Bar colors, the **Focused window** swatch is greyed while *App symbol style* is not *Glyphs* — a control on the Bars destination — and its only carrier was the dimmed row's hover string, which the owner could not raise (#1310).

The first cut put the reason on the card's header `?` (the Borders card's shape, and what `ui-designer` recommended). The owner looked at it on the device and asked whether that was understandable — it was not: a header `?` scopes the *card*, and with one greyed row inside a live card nothing points the glyph at the row. **Owner ruling 2026-09-14:** such a row takes a `CrossReferenceRow` directly beneath its grid, outside the dimmed subtree — the sentence states the fact and the destination name is the link, the per-space override editor's shape. The header `?` stays the shape for a greyed *block*.

- `SpaceBarColorCard` draws `AdvancedColorsHelp.focusedItemReference` under the at-rest grid while the bar is on and the picker is not Glyphs (`AdvancedColorsGates.focusedItemNeedsReference`); off, the header's bar-off reason covers the row as before.
- The sentence: *“Focused window” tints nothing while “App symbol style” is “System default” — set it to “Glyphs” in **Bars**.* — four labels by key (#818), the fifth slot the link; drafted in all ten catalogs through the scripts (localization-auditor), with the CJK and `ko` drafts reordering the condition first as their existing `*.xref` keys do.
- **The Borders card had the same shape one card up** — *Unfocused windows* greyed alone while the ring is on, its reason on the heading (the owner spotted that `?` next). Its header `?` now answers only for the ring being off, when every row is grey; unfocused-off draws `AdvancedColorsHelp.unfocusedReference` under the grid, linking Gaps & Borders. Also drafted in all ten catalogs.
- The row hover strings stay as the pointer fallback; the tooltip itself is not chased.
- `docs/design-decisions.md` ▸ *a dim is not a sentence* carries the amended ladder (channel 3's two shapes and which a row takes); `docs/ui-patterns.md` the convention — both version-less `:::unreleased`.

## Related Issue(s)

Fixes #1310

## Checklist

## How I verified

- verify-gate: build, `swift test -q` (5199 / 997 green), lint exit 0, site build, `unreleased-strip --check`; release build left to CI.
- guard-prover on the first cut, all red-proofed (the `ColorsGateTests` clauses, the `GreyOutAnchorTests` register, the `InterpolatedLabelTests` floor — proven to count the destination title). The re-cut keeps the same three guards with the new predicate, needle and floor, plus `CrossReferenceRowSlotTests` ▸ `theFocusedItemProsePlacesItsLink` and its `everyRowIsAsserted` coupling (which redded on the new row until registered); a prover pass on the re-cut is still owed before merge.
- code-reviewer on the first cut: 0 blockers, 0 major, 2 minor, applied; the re-cut moves the same sentence to the same class of site the reviewer had already checked (`SidebarCrossReferenceTests`, `CrossReferenceRowSlotTests`).
- Device: the branch's signed build is running on the dev Mac (65/65 apps attached). The owner saw the Space Bar link row on the previous build of this branch and ruled the Borders card into the same shape; the Borders half is on the running build. **Owner eyeball of both link rows still owed** before auto-merge is armed: Settings ▸ Advanced Colors — under *Border colors* and under *Space Bar colors*, one sentence each, the destination name a link.
- Measured, inconclusive: a synthetic 4 s hover on the greyed row on the previous build showed no tooltip; the positive control was disrupted by the owner using the machine, so the `GreyOut` "AppKit delay until proven otherwise" ruling stands unre-measured. Not needed for this fix.

## Notes for Reviewer

The `SettingKey` census entry for the row is unchanged — its `gate:` data still matches the predicate; the header `?` is the existing `SettingsSection(help:)` affordance, not a new control.


hajiboy95:
guard-prover on the re-cut (HEAD 869ddfc1): all four guards red-proofed — the three `focusedItemNeedsReference` clauses (each mutation reaches a different clause), the `GreyOutAnchorTests` needle (removing the row also reds `CrossReferenceRowSlotTests` ▸ `everyRowIsAsserted`), the `InterpolatedLabelTests` floor of 4, and `theFocusedItemProsePlacesItsLink` — which the prover showed is the ONLY net for a dangling sentence: neither lint nor `extract-keys --check` sees an argument swap. Stated limit: the anchor needle counts the expression, not its container, so a row moved inside the dimmed subtree would stay green; review's.

Issue #1414, opened 2026-09-13
🐛 A third-party window closed and reopened in a scrolling Space is bounced by the placement distrust


hajiboy95:
### KiwiDesk version or commit

`main` at 0a845d20 with PR #1412 (the #1380 fix) applied; observed by code reading during that review, not yet on device.

### macOS version

macOS 26.6.2 (25G83) — the dev machine; the class is version-independent.

### Mac model type

Apple Silicon (M1, M2, M3, etc.) — M1 Max

### Steps to reproduce

1. Put a Space into **scrolling** mode.
2. In any AppKit app whose secondary window survives a close (`isReleasedWhenClosed = false` — a Preferences/Settings window opened by ⌘, is the usual shape), open that window, then close it with its close button.
3. With another app's window focused in the same Space, reopen that window from the first app's menu bar (⌘, from the app menu, or its Window menu).

### Expected vs. actual behavior

Expected: the reopened window takes focus.

Predicted actual (unmeasured): the window appears but focus snaps back to the previously focused window for up to two seconds — the #1161 placement distrust. The mechanism is the one #1380 measured for KiwiDesk's own Settings window, and nothing in it is KiwiDesk-specific: a re-shown `NSWindow` keeps its window number, so the destroy fold's `.departed` memory makes the arrival a RETURN that steals no focus (#636), the arrival retile stamps the placement ledger, and the app's own clickless make-key report is bounced in the active scrolling Space. #1380's fix is the own-window door (`focusOwnWindow`), which a third-party app cannot call.

### Relevant output and logs

None yet — this is the sibling class the #1412 architect review surfaced (question e), filed so it gets a device sitting rather than being lost. The signature to look for in the unified log (`subsystem == "com.kiwicanopy.kiwidesk"`) is `gone: w<n> still hosted on shown space … — closed` followed within seconds by `focus: w<n> placement bounce distrusted; re-asserting w<m>` for the same `n` on the reopen. A first open of the window is never affected (the fold grants a NEW window the focus), nor is any non-scrolling mode (the bounce there needs an origin the window refused).

### Other window managers or AX tools running?

- [ ] Other window managers (e.g., Rectangle, yabai)
- [ ] Accessibility tools (e.g., Keyboard Maestro)
- [x] None that I know of

### Extra context

[...]

Issue #1345, opened 2026-09-08
🐛 A Desktop switch bounces back when the same app has a window on another Desktop


hajiboy95:
### KiwiDesk version or commit

1.2.0 — `8c1fc61` (the installed `/Applications/KiwiDesk.app`)

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

Measured on the dev machine 2026-09-08 with a continuous
recorder running (unified-log stream + `kiwidesk subscribe` +
a 5 Hz `get_state` sampler), after a 1 Hz poll missed the
transient entirely. Single screen, three KiwiDesk Spaces,
four macOS Desktops.

The setup needs **two windows of the same app on two different
Desktops**. Finder is the easy case:

1. Have a Finder window on Desktop 1 (here: "Downloads",
   `w955679`).
2. Send a second Finder window to another Desktop —
   `kiwidesk move_to_desktop 2` with "Papierkorb" (`w959472`)
   focused. It is now the only managed window on Desktop 2.
3. Swipe to Desktop 2 with the trackpad. Do not click.

### Expected vs. actual behavior

**Expected:** you arrive on Desktop 2 and stay there.

**Actual:** you are bounced straight back to Desktop 1, and with
two same-app windows the Desktop ping-pongs. The 5 Hz sampler
recorded six switches the user did not ask for, over ~30 s
(`desktop=` is the macOS Desktop, `s1f=` Space 1's focused slot):

```
+36.52s  desktop=2 active=1 s1f=959472 away=8
+40.08s  desktop=1 active=1 s1f=955679 away=1   <- bounced
+42.22s  desktop=2 active=1 s1f=959472 away=8
+44.82s  desktop=1 active=1 s1f=955679 away=1   <- bounced
+62.06s  desktop=2 active=1 s1f=959472 away=8
+64.10s  desktop=1 active=1 s1f=955679 away=1   <- bounced
+68.68s  desktop=2 active=1 s1f=959472 away=8
+69.10s  desktop=1 active=1 s1f=712721 away=1   <- bounced
```

The owner, watching it happen, described it as "somehow on
desktop change i got moved back to other desktop" and then
"somehow it bounced back to the finder in the current desktop.
so maybe because same application (even though different
window?" — which is the mechanism.

### Relevant output and logs

The streamed log carries the whole cycle, and it repeats
verbatim:

```
[...]

Issue #1281, opened 2026-09-06
🐛 Clicking the Settings window is bounced for two seconds after the row pans it out


hajiboy95:
### KiwiDesk version or commit

1.1.2, built from `e24fdfdf` (the tip of `main` after #1280), running from `.build/app`.

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

On a Space in **scrolling** mode, with enough windows that the Settings window scrolls out of view:

1. Open KiwiDesk Settings (menu bar ▸ Settings…). It scrolls into view and takes focus.
2. **Click** another window in the row. The row pans, and Settings scrolls out of view.
3. Within about two seconds, click the menu bar item again and pick **Settings…** (the App menu's ⌘, and the `open_settings` chord behave the same).
4. Repeat step 3 a few times, then wait four seconds and try once more.

The KiwiDesk focus verbs (a bound *focus left/right*) reach the scrolled-out Settings window at step 3 without any wait. Only the mouse path fails.

### Expected vs. actual behavior

**Expected:** Settings comes forward and the row pans onto it, as it does when a focus verb reaches it, whenever it was asked.

**Actual:** Nothing visible happens. Settings pops for an instant and focus snaps back to the window you were on. Each retry inside the window extends the lockout; after about four seconds of not trying, the same click works. Owner-reported 2026-09-06.

**Diagnosis, confirmed from the device log rather than hypothesised.** The open goes through `SettingsWindowController.show()`, which only calls `NSApp.forceFront`. The AX focus report for the Settings window therefore arrives as a CLICKLESS focus (the press landed on the status item, not on the marked window, so the #953 monitor records no provenance for it) of a window a pan placed less than two seconds earlier, in the active scrolling Space. That is exactly #1161's placement-bounce arm, and each distrust `renew`s the ledger, which is the "few seconds' distance".

The keyboard verbs work because they go through `KiwiCore.focusWindow`: state focus is set FIRST and a self raise stamped, so the report arrives with `intended == id` and is honored. The unified log of the reproduction (window 803284 is the Settings window, 712721 another app's window; the 13:39:00–01 lines are focus verbs, the 13:39:03 line is the click away, and the four distrusts are four menu clicks):

```
13:39:00.737 focus: w785748 (Antigravity IDE) honored; before w785748, selfEcho=true
[...]

Issue #958, opened 2026-08-24
🐛 Starting VoiceOver moves focus from the Settings window to the previous app's window


hajiboy95:
### KiwiDesk version or commit

0.9.9 dev, branch `feat/voiceover-hardening` at `6a37d4a1` (also reproduces on the merged `main` at `97380bc9` — the lane changed nothing in Core).

### macOS version

macOS 26.6.2

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Have a tiled Space with the KiwiDesk Settings window and another app's window (Zen browser here) side by side.
2. Click into Settings so it is the focused window.
3. Turn VoiceOver on (⌘F5).
4. Watch which window is focused a few seconds later.

### Expected vs. actual behavior

Expected: Settings keeps focus — starting a system accessibility service is not a focus change the user asked for.

Actual: focus moves to the other window (Zen) 3–8 s after VoiceOver starts. Turning VoiceOver off and on again repeats it. Observed by the owner during the #812 device sessions (2026-08-24).

### Relevant output and logs

Unified log (`subsystem == "com.kiwicanopy.kiwidesk"`) around one VoiceOver start; `w77602` is the Settings window, `w260` is Zen:

```
02:01:01.377 activation: untracked w77609 (com.apple.universalaccesscontrol) flagged ignored panel
02:01:01.383 focus: untracked w77609 (com.apple.universalaccesscontrol) flagged ignored panel
02:01:04.079 focus: w260 starts ignored-panel dismiss grace [1260, 32900]
02:01:04.079 focus: w260 (Zen) honored; before w77602, selfEcho=false
02:01:05.879 focus: w77602 cleared stale ignored-panel flags [1260, 32900]
02:01:05.879 focus: w77602 (KiwiDesk) honored; before w260, selfEcho=false
```

Same shape on a second start at 02:02:11 → 02:02:19. KiwiDesk *honors* the move rather than initiating it, so the mechanism is upstream of the focus handler: the system's accessibility control process (`com.apple.universalaccesscontrol`) takes activation, and when it yields, activation lands on the previous app instead of returning to Settings. That is the #952 activation-yield family (an own window losing activation to a transient system process) with a new trigger, and the same ignored-panel distrust machinery from #951 is in the trace.

### Other window managers or AX tools running?

- [x] None that I know of

(VoiceOver itself is the trigger; nothing else runs beside KiwiDesk.)

### Extra context

[...]

PR #1550, opened 2026-09-21
fix(focus): a menu-bar reveal's activation is returned to the own window (#1532)

### LINKED ISSUE #1532: 🐛 macOS 27: revealing the auto-hidden menu bar over a focused Settings window moves focus to the previously focused window

1.4.0 (d2b411d), the packaged build; repo HEAD ebaada16.

### macOS version

macOS 27.0 (26A428)

### Mac model type

### Steps to reproduce

1. System Settings ▸ Control Center ▸ *Automatically hide and show the menu bar* → Always.
2. Open KiwiDesk Settings and click into it so it is the key window.
3. Move the pointer to the top edge of the screen so the menu bar reveals.

**Expected:** the menu bar reveals and focus stays on the Settings window. (On macOS 26 the bar did not reveal at all while a KiwiDesk window was active — the `docs/accepted-limitations.md` row from #329: an `.accessory` app owns no menu bar. On 27 it reveals, which is the better half.)

**Actual:** the menu bar reveals **and focus moves to the window that was focused before Settings**. In a bsp Space it moves at once. In a scrolling Space it takes a few reveals before it moves — which smells like KiwiDesk's own scrolling-Space guard, the placement bounce (#1161: a clickless focus report in the active scrolling Space is bounced for two seconds), absorbing the first ones.

**Measure before fixing** (memory: quote the log lines that prove the mechanism). The likely OS behaviour is that macOS 27 reveals the menu bar of the last *regular* app and activates it, since the accessory app in front has none — a clickless, OS-driven activation with a cmd-tab's shape, the same class as #958's VoiceOver handoff (`KiwiCore+AccessibilityReturn`, a one-shot return) and #1322's provenance question. Log the focus reports around a reveal with the app the report came from (`reportsFromActiveApp` / `frontmostPID`) and `ownKeyWindow`, in bsp and in scrolling, then decide: return the focus (an `AccessibilityReturn`-shaped one-shot, armed while an own window is key and a reveal is in progress), or accept it and re-word the limitation. Either way the #329 row (a macOS 26.6.2 observation) is stale on 27 and moves in the same change.

### Relevant output and logs

To capture on the device: `log stream` filtered by KiwiDesk's pid during a reveal, in a bsp Space and in a scrolling Space, plus `get_state` before and after. No log attached yet — the report is the owner's eyeball on 27.0.

- [x] None that I know of

### Extra context

New since macOS 27; the same steps on 26 revealed nothing. #1386 (bar strip after auto-hide is turned off) is the geometry side of the same OS setting. Related rulings: *Permanent accessory mode (no activation policy switching)*, *A placement bounce is the app's answer, not the user's (#1161)*.

### Ruling (owner, 2026-09-21)

**Return the focus, once per reveal.** A clickless focus report from another app, arriving while the focus anchor is a window of KiwiDesk's own pid, the menu bar auto-hides and the pointer sits in the hidden strip, is macOS activating the last regular app so it has a bar to reveal — not the user leaving Settings. Core keeps state focus on the own window and re-asserts it with the STAMPED raise (`raiseWindow`, which mints the self stamp — #1281's point), so the report that follows is our own echo and the scrolling placement distrust does not bounce it; never the focus command, whose displacement note would arm that same bounce against every later report from the app (review, 2026-09-21). The #958 shape one arm over: evaluated at the report from readable facts rather than armed ahead, gated on `!selfEcho`, stood down by any left press since the reveal (a click into the revealed bar's menus is the user choosing that app), never when the re-assert would switch Desktops (#1345), and bounded to one return per short window so nothing can fight a second activation. The accepted trade, stated once: a deliberate cmd-tab away from an own window with the pointer parked at the top edge is returned once; the next one goes through.

The `docs/accepted-limitations.md` row from #329 (a macOS 26.6.2 observation) is stale on 27 and moves in the same change: on 27 the bar reveals — the previous app's bar, above the still-key own window — and focus stays.

hajiboy95:
## Description

With the menu bar set to auto-hide, macOS 27 answers the pointer reaching the top edge by activating the last *regular* app so it has a bar to reveal — KiwiDesk is an accessory app and owns none — and the clickless focus report that follows moved focus off the Settings window. Measured on the device: the activation lands ~190 ms after the pointer reaches the edge, ahead of the bar itself, and re-activating KiwiDesk while the pointer stays there holds (macOS does not steal it back; the previous app's bar stays revealed above the key own window and hides when the pointer leaves).

The fix is the #958 return one arm over, decided at the report from readable facts rather than armed ahead: a report from another app while the focus anchor is a window of our own pid, no left press inside the echo window (a click into the revealed bar's menus is the user choosing that app), the pointer in the reveal strip, no Desktop switch (#1345). State focus stays on the own window and is re-asserted with the **stamped** raise (`raiseWindow`), so the report that follows is our own echo (#1281's point) and the scrolling placement distrust does not bounce it — an unstamped activation was measured landing in exactly that bounce (#1414's class on the own window). Not the focus command either: its displacement note would put the foreign window in the placement ledger, and in an active scrolling Space that bounces every later report for the ledger's window, so the one-shot bound would never decide there (code review). One return per echo window, so an activation KiwiDesk cannot hold is never fought twice. The strip read lives behind `mouse.pointerInMenuBarStrip`, pinned false by both `makeTestCore` twins.

The #329 accepted-limitations row was a macOS 26 observation; it now says what each OS does, with the return in a version-less `:::unreleased` note. The ruling is on the issue and in design-decisions.

## Related Issue(s)

Closes #1532

## Checklist

  (see AGENTS.md §3)
  required)
- [x] I have signed the
  [CLA](https://github.com/KiwiCanopy/KiwiDesk/blob/main/CLA.md)
  (the bot asks on your first PR)

## How I verified

- Gate: `swift build`, `swift test -q` (1606 + 4021 tests green), `scripts/lint.sh`, and the site build (`docs/` changed). Release build left to CI — no concurrency boundary changed.
- Mechanism, on the unfixed 1.4.0 build (macOS 27.0, auto-hide on): a probe warps the pointer to the top edge with Settings key and polls `frontmostApplication` and the on-screen menu-bar window every 30 ms — Claude activated at ~190 ms, its bar appeared at ~310 ms, KiwiDesk logged `focus: w14897 (Claude) honored`. Re-activating KiwiDesk 1.5 s later with the pointer still at the edge held for the 3 s the pointer stayed (bsp); in scrolling the same re-activation was bounced by the placement distrust, which is why the return takes the focus command. The measurement tables are on #1532.
- Fixed build on the device: the packaged bundle relaunched, then the same probe — the log line `focus: w<n> menu-bar reveal activation returned; re-focusing w<own> (#1532)` and KiwiDesk frontmost after the reveal, in bsp and in scrolling. (Pending the owner's Accessibility re-grant at the time of opening; the result lands in a comment.)

## Notes for Reviewer

- The trims in `KiwiCore+FocusEvents.swift` and `KiwiCore.swift` are comment-only, to stay under the 350-line ceiling; the arguments they carried live in the owning files and design-decisions.
- The reveal band is the deepest of the screen's safe area, its reserved top strip and the installed main menu's bar height — a bare `NSMenu()` answers 0 (measured), and `NSStatusBar` is censused by the status-item seam guards.
- The live strip read is gated to macOS 27+: on 26 the reveal activates nothing (the #329 row), so the arm would be all cost there.


hajiboy95:
### Device proof, fixed build (5aa04245 packaged, macOS 27.0, auto-hide on, Settings key)

The same probe as the measurement on #1532: pointer warped to the top edge, `frontmostApplication` and the on-screen menu-bar window polled every 30 ms, KiwiDesk's log streamed by pid.

**Scrolling Space** — KiwiDesk never left the front at the poll's resolution; the bar still revealed:

| Event | After the pointer reached the edge |
|---|---|
| `focus: w14897 menu-bar reveal activation returned; re-focusing w62008 (#1532)` | 165 ms |
| `focus: w62008 (KiwiDesk) honored; before w62008, selfEcho=true` (our raise's echo) | 211 ms |
| the previous app's bar on screen, KiwiDesk frontmost | 285 ms |

**bsp Space** — the activation is visible for one poll tick, then returned:

| Event | After the pointer reached the edge |
|---|---|
| Claude frontmost | 175 ms |
| the return logged | 176 ms |
| KiwiDesk frontmost again | 231 ms |
| the previous app's bar on screen, KiwiDesk frontmost | 289 ms |

No second activation from macOS in the 5 s the pointer stayed at the edge in either mode; the bar hid when the pointer left with KiwiDesk still frontmost.

Note for the eyeball: the revealed bar is the previous app's (its menus and the status items), above the still-key Settings window — that is the OS's choice and the ruling accepts it.

PR #1556, opened 2026-09-21
fix(focus): a window returning from a close takes the focus at its arrival (#1414)

hajiboy95:
## Description

An `NSWindow` keeps its number across a close and a re-show, so a window the user closed and reopened came back to Core as a RETURN that steals no focus (#636). In an active scrolling Space the arrival retile placed it and the app's own make-key report was bounced by the #1161 placement distrust — measured on the device with Telegram's main window, 3 of 3 in scrolling, 0 in bsp (the tables are on #1414).

The provenance now moves to where it is known: the gone handler already classifies a close apart from a Desktop departure and a hide, so its `closed` arm marks the departure (`rememberClosedDeparture`), and the create fold consumes the mark on every arrival and grants a close-return the focus a new window gets, in the active Space only. The report then lands intended, which the predicate never reads — no exemption in it. A Desktop `vanished` keeps #636's rule, a hide #913's; the mark rides the re-key and the away retire with the departed memory.

The #1380 own-window debt (`ownShowFocus`, `payOwnShowFocus`, `OwnShowFocusSeamTests`) is retired: the fold grants what it paid, and the door answers an untracked number true without owing anything. `DesktopFocusPaymentTests`' fixture now flips its topology before the destroys — the ordering #1207's trace measured — so its departures classify `vanished` rather than as closes.

## Related Issue(s)

Closes #1414

## Checklist

- [x] I understand what this change does and have run it in
  the app to confirm it works as intended (see "How I
  verified" below and CONTRIBUTING.md → Using AI Assistants)
- [x] Commits follow Conventional Commits format
  (see AGENTS.md §3)
- [x] New behavior is tested (pure logic / layout code
  required)
- [x] User-facing changes are documented in `docs/`
- [x] No contradictions between code and docs
- [x] PR is focused; refactors separated from features
- [x] I have signed the
  [CLA](https://github.com/KiwiCanopy/KiwiDesk/blob/main/CLA.md)
  (the bot asks on your first PR)

## How I verified

- Gate: `swift build`, `swift test -q` (1603 + 4031 green), `scripts/lint.sh`, `scripts/unreleased-strip --check`, and the site build (`docs/` changed). Release build left to CI — no concurrency boundary changed.
[...]

Issue #913, opened 2026-08-21
🐛 A hidden app keeps its tiles — Discord's red X hides the app, so its window never leaves the register


hajiboy95:
### KiwiDesk version or commit

0.9.8 (fe17b18)

### macOS version

macOS 26.6.2 (25G83)

### Mac model type

Apple Silicon (M1, M2, M3, etc.)

### Steps to reproduce

1. Open Discord and let it tile (App Bar shows its item, and it
   holds a layout slot).
2. Click the window's red **X**.
3. Watch the App Bar.

The general form, which needs no Discord at all:

1. Focus any tiled app.
2. Press **⌘H**.
3. Watch the App Bar.

### Expected vs. actual behavior

Expected: the window leaves the register and the layout closes
over the gap, the way it does on minimize.

Actual: the item stays in the App Bar and the window keeps its
layout slot — indefinitely, until the app quits. The tile is held
for a window that is on screen nowhere.

Minimize is correct today; only hide is affected.

### Relevant output and logs

Nothing is logged, and that is itself the finding: from
KiwiDesk's point of view nothing happened, so there is no error
to report. The evidence is a probe sampling `get_state` against
the WindowServer once a second across all three gestures:

```
21:17:01  minimized   AX min=true    onscreen=false   register: 0 windows  OK
21:17:11  reopened    AX min=false   onscreen=true    register: 1 window   OK
21:17:19  pressed X   AX min=false   onscreen=false   register: 1 window   STALE
```

## The mechanism

Discord's X does not close a window — it hides the whole
application:

```
NSRunningApplication(pid 56441): bundle=com.hnc.Discord isHidden=true
AX application element:          AXHidden = 1
AX window list:                  count=1, AXMinimized=0, title="@cal - Discord"
```

That is the same state as ⌘H, and Electron apps that live in the
background do it when their last window closes.

`EventLoop.reconcile` decides a window is gone by asking the app
for its AX window list and treating anything missing from it as
closed. A window leaves that list two ways: it is destroyed, or
`AXMinimized` is true. A hidden app is neither — AX keeps
reporting the window, un-minimized, at its old position and size
— so reconcile looks, sees a healthy live window, and keeps the
slot. Minimize works precisely because it is the one gesture that
flips `AXMinimized`.

The reconcile does run: `appActivated` reconciles the app you
just left. It runs and concludes nothing changed.

Nothing in the tree observes hide/unhide —
`didHideApplicationNotification` and
[...]

PR #1562, opened 2026-09-21
feat(focus): a window you bring back lands where you are, like a new window (#1561)

### LINKED ISSUE #1561: 💡 A window you bring back — closed and re-shown, or minimized and restored — lands where you are, like a new window
### Use case

Bringing a window back is inconsistent today. A window you **closed** in an app whose close hides it (Telegram's main window, KiwiDesk's own Settings — the same `NSWindow` comes back under the same number) returns to the Space it was closed in, with its old slot and track break, and the focus follow then switches you there. A window you **minimized** and restore from the Dock lands as if new: where you are, or where its app rule says. The destroy fold files the Space in the departed memory on a close and clears it on a minimize — no ruling argued either side (session 2026-09-21, surfaced by the #1414 eyeball).

### Proposed solution

Treat a window you bring back — a close re-shown, a minimize restored — the way a NEW window is treated: its app rule if it has one, otherwise the Space you are on. It never goes back to its old Space on its own. The close mark #1414 introduced already tells a close apart from a Desktop departure at the return, so the fold can route a marked return through the new-window target instead of the remembered one; the minimize side already does this.

Kept as they are: a window that LEFT WITH ITS DESKTOP (#1207) and a window of an app hidden with ⌘H (#913) — those come back as a burst to the Spaces they were in, which is what macOS does and what the user did not ask to change.

### Alternatives considered

Making the minimize side return to its Space too, like macOS does for both (keeps the rank-and-break restore for minimized windows; restoring from the Dock could then switch your Space). Rejected by the owner: a window you bring back is wanted where you are.

### Extra context

Trade the ruling accepts: a re-shown closed window no longer takes back its slot and track break in the Space it left; it joins the active Space's row like a newcomer. The #1414 focus grant is unaffected — the return still takes the focus at its arrival, now in the Space it lands in. Builds on PR #1556's close mark, so it lands after it.

### Ruling (owner, 2026-09-21)

**A window you bring back lands where you are.** A close re-shown and a minimize restored take a new window's target — an app rule where one names the app, otherwise the active Space — never the Space they left. "I need it now, where I am" is the intuition; the rank-and-break restore for a closed window is given up for it. A Desktop return and a ⌘H unhide keep their rules: those are the OS bringing a burst back, not the user bringing one window back.

**For the release notes (owner, 2026-09-21):** say this is a deliberate choice, not a fix — a window manager opens windows where you are, so you can move them somewhere else if you want; it does not send them back to where they once were.

hajiboy95:
## Description

A window you closed in an app whose close hides it (Telegram's main window, KiwiDesk's own Settings) came back to the Space it was closed in, with its old slot and track break, and the focus follow then switched you there — while a minimized window restored from the Dock landed as new, where you are. Owner ruling (2026-09-21, on #1561): a window you bring back is wanted where you are, so you can move it somewhere else if you want; a window manager opens windows where you are rather than sending them back to where they once were.

The create fold, reading the close mark #1414 introduced, now drops the departed memory and the slot record and takes the new-window path: the app rule, else the active Space, with the focus a new window gets there. The #1414 focus grant folds into that arm. A Desktop return (#1207) and a ⌘H unhide (#913) keep their rules — those are the OS bringing a burst back, not you bringing one window back.

**For the release notes (owner, verbatim):** "in the changelog later we should also say that we deliberately made this choice to do so since that's like a window manager should work. open windows where you are such that you can move them somewhere else if you want."

## Related Issue(s)

Closes #1561

## Checklist

  (see AGENTS.md §3)
  required)
- [x] I have signed the
  [CLA](https://github.com/KiwiCanopy/KiwiDesk/blob/main/CLA.md)
  (the bot asks on your first PR)

## How I verified

- Gate: `swift build`, `swift test -q` (1603 + 4035 green), `scripts/lint.sh`, `scripts/unreleased-strip --check`. Release build left to CI — no concurrency boundary changed.
- Tests: `ClosedReturnFocusTests` gains the placement half — lands where you are, follows the app rule, gives up its old slot, outranks a restore filed after the close; `PlacementIntentTests` re-points the off-Space case to the new rule. Reviews (code, architecture, docs) and a prover pass ran on the first commit; the second commit answers them.
- Device: the packaged build relaunched, Telegram's window closed in one Space and reopened from another — it lands in the Space you are on, focused. The result lands in a comment below.

## Notes for Reviewer

- The trade: a re-shown closed window no longer takes back its slot and track break; the design entry and the limitations note say so.
- Stacked on #1556's close mark; that PR merged first, so this one is rebased onto its squash.


hajiboy95:
### Device proof (8e4c9463 packaged, macOS 27.0)

Telegram's main window (`w91435`) shown in Space 1 (scrolling), closed through its AX close button, `kiwidesk focus_space 2`, then the `reopen` Apple event plus `activate` (the Dock click's shape):

| Log line | |
|---|---|
| `gone: w91435 still hosted on shown space 1 — closed` | the close, marked |
| `close return: w91435 re-shown — placed as new, focus taken (#1414/#1561)` | the fold's new-window placement |
| `focus: w91435 (Telegram) honored; before w91435, selfEcho=false` | the app's report, intended |

`get_state` after the reopen: active Space 2, Telegram listed in Space 2 and focused there; Space 1's row unchanged, its focus still on the window it had. Telegram frontmost for the 3 s polled. Before this change the same sequence put the window back into Space 1 and the focus follow switched the user there.

